PR #6, opened 2022-08-11
Add multiple Sessions support

### LINKED ISSUE #3: Webview freezes when modal closed
### Description

When the `webview` screen with `presentation: "modal"` is closed, the screen underneath doesn't gain focus.

### Why?

`visitableWillAppear` in the `RNVistableView` is not called, see [why here](https://hacknicity.medium.com/view-controller-presentation-changes-in-ios-13-ac8c901ebc4e).

Fryzu:
Fixes #3 #8 by introducing support for multiple sessions.

The PR adds new native component `RNSession.swift` which stores instance of the Turbo Session. The id of the session is passed to all its React children via `React.Context` (SessionContext.ts) and consumed automatically by the VistiableView.tsx.


ios/RNVisitableView.swift
```diff
@@ -14,17 +14,34 @@ class RNVisitableView: UIView {
...
+  required init?(coder aDecoder: NSCoder) {
+    super.init(coder: aDecoder)
+  }
+  
+  public func getSession() -> Session? {
+    guard let view = self.bridge?.uiManager?.view(forReactTag: sessionHandle!) as? RNSession else {
+      print("Couldn't find session for sessionHandle:", sessionHandle!)
+      return nil
+    }
+    return view.session
```
lukmccall, line 36:
```suggestion
    guard let sessionHandle = sessionHandle else {
      print("Couldn't find session for nil")
      return nil
    }
    
    guard let view = self.bridge?.uiManager?.view(forReactTag: sessionHandle) as? RNSession else {
      print("Couldn't find session for sessionHandle:", sessionHandle)
      return nil
    }
    return view.session
```

lukmccall: Also, I would add a check if that function was called on the main thread - you can use `Thread.isMainThread` to do it.

ios/RNSession.swift
```diff
@@ -0,0 +1,18 @@
...
+
+import Turbo
+import UIKit
+
+class RNSession: UIView {
+  
+  public var session: Session = {
+      let session = Session()
+      return session
+  }()
```
Fryzu, line 16:
```suggestion
  public var session = Session()
```

ios/RNVisitableView.swift
Fryzu: Hello @lukmccall thanks for review. I was thinking about `Thread.isMainThread` and `caching` issue 🤔 but still don't really understand the issue. Can you add some more context?

Fryzu: Resolved as discussed.

Issue #13, opened 2022-09-15
Multiple Session support for Android


Fryzu:
### Description
Right now every visitable view uses same `Session`.  
### Why?
Right now we don't use any advantages of turbo.

PR #15, opened 2022-09-23
@fryzu/android multiple session

Fryzu:
## Adds support for multiple session on Android

[WIP] It sometimes does not attach the webview, currently working on it.

Fixes: https://github.com/software-mansion-labs/react-native-turbo-demo/issues/13

### DIFF
--- a/android/app/src/main/java/com/hotwirerndemo/RNSession.kt
+++ b/android/app/src/main/java/com/hotwirerndemo/RNSession.kt
@@ -5,6 +5,34 @@ import android.widget.FrameLayout
 import androidx.appcompat.app.AppCompatActivity
 import com.facebook.react.bridge.ReactContext
 import dev.hotwire.turbo.session.TurboSession
+import dev.hotwire.turbo.views.TurboView
 import dev.hotwire.turbo.views.TurboWebView
+import dev.hotwire.turbo.visit.TurboVisit
 
-class RNSession(context: Context) : FrameLayout(context)
+class RNSession(context: Context) : FrameLayout(context) {
+
+    lateinit var session: TurboSession private set
+    private val reactContext = context as ReactContext
+    private var prevTurboView: TurboView? = null // Needed to webView when changing view
+
+    init {
+        setupNewSession()
+    }
+
+    private fun setupNewSession() {
+        val activity = reactContext.currentActivity as AppCompatActivity
+        val webView = TurboWebView(context, null)
+
+        session = TurboSession("testSessionName", activity, webView)
+        session.setDebugLoggingEnabled(true) // TODO, remove
+    }
+
+    internal fun visit(visit: TurboVisit, view: TurboView) {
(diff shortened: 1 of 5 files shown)


ios/RNVisitableView.swift
```diff
@@ -126,4 +126,5 @@ extension RNVisitableView: SessionDelegate {
       // Handle non-Turbo links
   }
   
+
```
pfeiffer, line 129: Unneeded change

android/app/src/main/java/com/hotwirerndemo/RNVisitableView.kt
```diff
@@ -60,91 +79,63 @@ class RNVisitableView (context: Context) : LinearLayout(context), TurboSessionCa
         super.onLayout(changed, l, t, r, b)
         val width = r - l
         val height = b - t
-        session.webView.layout(0, 0, width, height)
+        session.session.webView.layout(0, 0, width, height)
```
pfeiffer, line 82: Minor detail but `session.session` reads pretty bad. You could consider renaming the `session` property in `RNSession` to eg. `turboSession` to be explicit about the session type.

android/app/src/main/java/com/hotwirerndemo/RNSession.kt
```diff
@@ -5,6 +5,58 @@ import android.widget.FrameLayout
...
+
+    init {
+        setupNewSession()
+    }
+
+    private fun setupNewSession() {
+        val activity = reactContext.currentActivity as AppCompatActivity
+        val webView = TurboWebView(context, null)
+
+        session = TurboSession("testSessionName", activity, webView)
```
pfeiffer, line 26: Should the TurboSession name also be dynamic and based on the RNSession name?

android/app/src/main/java/com/hotwirerndemo/SessionSubscriber.kt
```diff
@@ -0,0 +1,12 @@
...
+import dev.hotwire.turbo.visit.TurboVisit
+
+interface SessionSubscriber {
+
+    val visit: TurboVisit
+
+    fun detachWebView(callback: () -> Unit)
+    fun attachWebView()
+
+}
```
lukmccall, line 12:
```suggestion
}

```

android/app/src/main/java/com/hotwirerndemo/RNSession.kt
Fryzu: > The name of the TurboSession (this is arbitrary, but must be unique in your app)
^ From docs ^
I will make it unique then.

android/app/src/main/java/com/hotwirerndemo/RNVisitableView.kt
Fryzu: 👀 It will be fixed in the following PR

PR #20, opened 2022-11-17
Support pushing wildcard screens; fix `replace` action

pfeiffer:
This PR fixes an issue with pushing the wildcard screen to the stack. Previously we dispatched a navigation action with `type: "NAVIGATE`. The behavior of the navigator is to *not* push the same screen multiple times, even when the params have changed.

We should instead use the `PUSH` action from the StackNavigator.

This PR also fixes an issue where previously a Turbo action of type `replace` **would navigate back** instead of replacing the current screen with the new one.

### DIFF
--- a/packages/navigation/src/useWebviewNavigate.ts
+++ b/packages/navigation/src/useWebviewNavigate.ts
@@ -6,7 +6,21 @@ import {
 import * as React from 'react';
 import LinkingContext from '@react-navigation/native/src/LinkingContext';
 import extractPathFromURL from '@react-navigation/native/src/extractPathFromURL';
+import { StackActions } from '@react-navigation/native';
+import type {
+  NavigationState,
+  NavigatorScreenParams,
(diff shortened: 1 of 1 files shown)


packages/navigation/src/useWebviewNavigate.ts
```diff
@@ -6,8 +6,22 @@ import {
...
 import LinkingContext from '@react-navigation/native/src/LinkingContext';
 import extractPathFromURL from '@react-navigation/native/src/extractPathFromURL';
+import { StackActions } from '@react-navigation/native';
+import type {
+  NavigationState,
+  NavigatorScreenParams,
+} from '@react-navigation/core';
 import type { Action } from '@react-native-turbo-webview/core';
 
+type NavigateAction<State extends NavigationState> = {
```
pfeiffer, line 16: This type isn't exported by `react-navigation`, so need to copy it here. Hm.

PR #23, opened 2022-12-02
Add detach rewrite with coroutines

Fryzu:


Fryzu:
Hey @aleqsio I noticed issues with this branch, the empty screen appears event when screen is pushed on the navigation stack 🤔

PR #43, opened 2023-09-27
Fix onMessage not working on iOS

pklatka:
## Summary

This PR:
1. Fixes an issue with `onMessage` not working on iOS. The main issue was that on Android, when a message is sent from the website, the event is being sent and caught with a `DeviceEventEmitter` listener. However, on iOS, the event was not being sent at all.
2. Replaces `DeviceEventEmitter` with `NativeEventEmitter `because it is deprecated.
3. Checks for the existence of the `AndroidInterface` variable. When the app was loading the example website on iOS, screen sharing would not work because `AndroidInterface` is not defined on iOS.

## Test plan

Run the app on iOS device and check if messages from the website are shown in the console.

### DIFF
--- a/packages/turbo/ios/RNSessionModule.swift
+++ b/packages/turbo/ios/RNSessionModule.swift
@@ -12,6 +12,8 @@ class RNSessionModule: RCTEventEmitter {
   
   @objc
   private var sessions: [NSString: RNSession] = [:]
+    
+  @objc var supportedEventNames: [String] = []
   
   @objc
   private var defaultSession = RNSession()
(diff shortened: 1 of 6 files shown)


packages/turbo/ios/RNSessionModule.swift
```diff
@@ -30,6 +33,9 @@ class RNSessionModule: RCTEventEmitter {
     _ sessionHandle: NSString,
     resolver resolve: @escaping RCTPromiseResolveBlock,
     rejecter reject: @escaping RCTPromiseRejectBlock) -> Void {
+      if let index = supportedEventNames.firstIndex(of: "sessionMessage" + (sessionHandle as String)) {
+        supportedEventNames.remove(at: index)
+      }
```
pklatka, line 38: You might be wondering if that's a good approach... Well, from what I tested, `removeSession` is called rarely, so it's better to do `O(n)` search rather than converting `session.keys` every time `supportedEvents` is called (and `supportedEvents` method is called every time website sends an event to native.

Fryzu (approved):

PR #44, opened 2023-09-27
Make user-agent of WebView customisable 

### LINKED ISSUE #7: Make user-agent of WebView match Turbo Native's
The Turbo Native WebViews includes "Turbo Native iOS" ([source](https://github.com/hotwired/turbo-ios/blob/4b313588654acf62675b9870cf780b2300585cbf/Demo/SceneController.swift#L50)) or "Turbo Native Android" ([source](https://github.com/hotwired/turbo-android/blob/c845dbc4b2845e730632b815368de7987201f1aa/demo/src/main/kotlin/dev/hotwire/turbo/demo/main/MainSessionNavHostFragment.kt#L57)) in the user-agent used by the webviews.

The turbo-rails integration uses these to determine how to handle special navigation cases:

```ruby
# Turbo Native applications are identified by having the string "Turbo Native" as part of their user agent.
def turbo_native_app?
  request.user_agent.to_s.match?(/Turbo Native/)
end
```

I believe this library should match that behavior, adding `Turbo Native [iOS|Android]` to the user-agent string to match behavior of the native libraries.

pklatka:
This PR is dependent on #53. Because of this reason, please review last three commits only!

## Summary

This PR resolves #7: it adds support for custom user-agent in `Session` and `VisitableView` components.

## Test plan

Run the app where custom user-agent is being handled and check if there are any differences when using custom user-agent string.

### DIFF
--- a/packages/turbo/ios/RNVisitableViewModule.swift
+++ b/packages/turbo/ios/RNVisitableViewModule.swift
@@ -21,10 +21,11 @@ class RNVisitableViewModule: RCTEventEmitter {
   @objc
   public func setConfiguration(
     _ sessionHandle: NSString,
+    applicationNameForUserAgent: NSString?,
     resolver resolve: @escaping RCTPromiseResolveBlock,
     rejecter reject: @escaping RCTPromiseRejectBlock) -> Void {
       if(sessions[sessionHandle] == nil){
-        sessions[sessionHandle] = RNSession(eventEmitter: self, sessionHandle: sessionHandle)
+        sessions[sessionHandle] = RNSession(eventEmitter: self, sessionHandle: sessionHandle, applicationNameForUserAgent: applicationNameForUserAgent)
         supportedEventNames.append("sessionMessage" + (sessionHandle as String))
       }
       resolve(sessionHandle)
--- a/packages/turbo/src/VisitableView.tsx
+++ b/packages/turbo/src/VisitableView.tsx
@@ -31,6 +31,7 @@ const RNVisitableViewModule = getNativeModule<VisitableViewModule>(
 export interface Props {
   url: string;
   sessionHandle?: string;
+  applicationNameForUserAgent?: string;
   onVisitProposal: (proposal: NativeSyntheticEvent<VisitProposal>) => void;
   onLoad?: (proposal: NativeSyntheticEvent<OnLoadEvent>) => void;
   onVisitError?: OnErrorCallback;
@@ -45,6 +46,7 @@ const VisitableView = React.forwardRef<RefObject, React.PropsWithRef<Props>>(
   (props, ref) => {
     const {
       sessionHandle = 'Default',
+      applicationNameForUserAgent,
       onMessage,
       onVisitError: viewErrorHandler,
     } = props;
@@ -52,10 +54,13 @@ const VisitableView = React.forwardRef<RefObject, React.PropsWithRef<Props>>(
 
     useEffect(() => {
       const setSessionConfiguration = async () => {
-        await RNVisitableViewModule.setConfiguration(sessionHandle);
+        await RNVisitableViewModule.setConfiguration(
+          sessionHandle,
+          applicationNameForUserAgent
+        );
       };
       setSessionConfiguration();
-    }, [sessionHandle]);
+    }, [sessionHandle, applicationNameForUserAgent]);
 
     useImperativeHandle(
       ref,
--- a/packages/turbo/README.md
+++ b/packages/turbo/README.md
@@ -78,6 +78,10 @@ URL for the WKWebview to open. Changing the url should result in view replacing 
 
 Session handle for the webview. If not provided, the default session will be used. It can be used to create separate webview instances for different parts of the app.
 
+### `applicationNameForUserAgent`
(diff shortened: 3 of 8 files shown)


packages/turbo/ios/RNVisitableView.swift
```diff
@@ -80,6 +81,7 @@ class RNVisitableView: UIView, SessionSubscriber {
 
 extension RNVisitableView: RNVisitableViewControllerDelegate {
   func visitableWillAppear(visitable: Visitable) {
+    getSession()?.webView.customUserAgent = applicationNameForUserAgent as String?
```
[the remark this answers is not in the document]
pfeiffer: This also seems to be the behavior on Android side, so let's ensure there's consistency :-) We can always decide to add a `userAgent` prop later on to customize the full user agent as needed.

packages/turbo/src/VisitableView.tsx
```diff
@@ -42,35 +44,44 @@ export interface RefObject {
...
+    const {
+      sessionHandle,
+      applicationNameForUserAgent,
+      onMessage,
+      onVisitError: viewErrorHandler,
+    } = props;
     const messageHandlerEventSubscription = useRef<EmitterSubscription>();
-    const { sessionHandle, onVisitError: sessionErrorHandler } =
-      useContext(SessionContext);
+    const resolvedSessionHandle = sessionHandle || 'Default';
```
sebryu, line 54: I would just put default value into props spreading operation ->
```
const { sessionHandle = 'Default', ...otherProsp } = props;
```

packages/turbo/ios/RNVisitableViewModule.swift
```diff
@@ -17,6 +17,18 @@ class RNSessionModule: RCTEventEmitter {
   
   @objc
   private var defaultSession = RNSession()
+    
+  @objc
+  public func setConfiguration(
+    _ sessionHandle: NSString,
+    resolver resolve: @escaping RCTPromiseResolveBlock,
+    rejecter reject: @escaping RCTPromiseRejectBlock) -> Void {
```
sebryu, line 25: @pklatka currently we are not using `reject`, right? Should we add it later at some point? Or we don't need custom logic here?

packages/turbo/android/src/main/java/com/reactnativeturbowebview/RNSessionModule.kt
```diff
@@ -8,10 +8,10 @@ import com.facebook.react.bridge.ReactMethod
 import com.facebook.react.module.annotations.ReactModule
 import java.util.*
 
-private const val MODULE_NAME = "RNSessionModule"
+private const val MODULE_NAME = "RNVisitableViewModule"
```
sebryu, line 11: I guess you should also rename the file 😉

packages/turbo/src/Session.tsx
```diff
@@ -1,100 +0,0 @@
...
-export interface Props {
-  onMessage?: SessionMessageCallback;
-  onVisitError?: OnErrorCallback;
-}
-
-interface State {
-  sessionHandle?: string | null;
-}
-
-export default class Session extends React.Component<Props, State> {
```
sebryu, line 26: as discussed - for better developer experience I would keep this file (until v1.0) that would put `console.error` + throw an error with link to this PR description that would explain how to change existing codebases (for those few individuals who uses our lib)

example/src/WebView.tsx
```diff
@@ -0,0 +1,58 @@
...
+}
+
+const sessionHandle = 'TurboWebviewExample';
+
+const WebView: React.FC<Props> = (props) => {
+  const { navigation, route } = props;
+  const navigateTo = useWebviewNavigate();
+  const path = route?.params?.path || route?.path;
+  const baseURL = route?.params?.baseURL || props?.baseURL;
+  const currentUrl = path ? `${baseURL}${path}` : baseURL;
```
sebryu, line 25:
```suggestion
  const path = route?.params?.path || route?.path || '';
  const baseURL = route?.params?.baseURL || props?.baseURL;
  const currentUrl = `${baseURL}${path}`;
```

sebryu (changes requested):
Looks good, biggest concern is already raised by Mattias, so please address that before merging 👏

packages/turbo/ios/RNVisitableViewModule.swift
pklatka: Hmm, we might use reject when we add something to RNSession that could potentially block us from using, for example, a webview, but for now, we don't need it.

packages/turbo/ios/RNVisitableView.swift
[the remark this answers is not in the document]
pklatka: BTW, Android unfortunately doesn't have something like `applicationNameForUserAgent` prop on WebView, instead you have to get the current user agent and then append custom string :) (see [here](https://github.com/react-native-webview/react-native-webview/blob/master/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManagerImpl.kt#L231-L257))

pfeiffer:
LGTM! 🎉

PR #47, opened 2023-11-05
Support urls with query string

pklatka:
This PR fixes the problem in `useWebviewNavigate` hook -> urls with query string were not properly handled, causing `pathWithScreenParams` variable to have improper format.

### DIFF
--- a/packages/navigation/src/useWebviewNavigate.ts
+++ b/packages/navigation/src/useWebviewNavigate.ts
@@ -74,10 +74,10 @@ export default function useWebviewNavigate<
 
       const { options } = linking;
 
-      const path =
-        (to.match(/^https?:\/\//)
-          ? extractPathFromURL(options?.prefixes, to)
-          : to) ?? '';
+      let path = to;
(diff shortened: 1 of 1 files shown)


packages/navigation/src/useWebviewNavigate.ts
```diff
@@ -61,13 +61,23 @@ export default function useWebviewNavigate<
...
+      const queryStringIndex = path.indexOf('?');
+      let pathWithoutQueryString, queryString;
+      if (queryStringIndex !== -1) {
+        pathWithoutQueryString = path.slice(0, queryStringIndex);
+        queryString = path.slice(queryStringIndex + 1);
+      } else {
+        pathWithoutQueryString = path;
+        queryString = '';
+      }
+      const pathWithScreenParams = `${pathWithoutQueryString}?${queryString}&path=${pathWithoutQueryString}`;
```
pfeiffer, line 80: Would it make sense to extract this into it's own utility method for better communicating what is going on?

PR #48, opened 2023-11-05
Restore scroll position

### LINKED ISSUE #4: Webview scrollview doesn't restore the scroll position
### Description

See the demo `Try Turbo's scroll restoration` page for more context.

pklatka:
This PR resolves #4. It sets visit option to `restore` the view.


packages/turbo/android/src/main/java/com/reactnativeturbowebview/RNVisitableView.kt
```diff
@@ -69,19 +69,9 @@ class RNVisitableView(context: Context, sessionModule: RNSessionModule) : Linear
   }
 
   fun visit(restoreWithCachedSnapshot: Boolean, reload: Boolean) {
-    val restore = !restoreWithCachedSnapshot && !reload
-
-    val options = when {
-      restore -> TurboVisitOptions(action = TurboVisitAction.RESTORE)
-      else -> TurboVisitOptions()
-    }
+    val options = TurboVisitOptions(action = TurboVisitAction.RESTORE)
```
pfeiffer, line 72: Wouldn't this make _all_ visits be initiated with the `restore` action, even normal advance visits? I'm thinking that we'd probably want to only perform restores for back navigations (where `restoreWithCachedSnapshot` is true and `reload` is false).

I think there might have been a logical error in the previous code, causing the `var restore = ..` to evaluate to `false` even when `restoreWithCachedSnapshot` is true which might be what was causing the scroll restoration issue.

Compare the code in the Turbo Android library with the one in this:

https://github.com/hotwired/turbo-android/blob/bf945e18aed3d5749c9efc6e228d98bc012878af/turbo/src/main/kotlin/dev/hotwire/turbo/delegates/TurboWebFragmentDelegate.kt#L346-L351


```diff
- val restore = restoreWithCachedSnapshot && !reload
+ val restore = !restoreWithCachedSnapshot && !reload
...
```

I think what we want is probably to copy the code from the Turbo Android library to mimick the behavior regarding reloads, advances and restores.

PR #50, opened 2023-11-05
Fix native ActivityIndicator not being shown on Android

pklatka:
On Android native progressView/ActivityIndicator was not visible when opening slow loading page. It was caused by react-native behavior: to achieve performance, not all native components are being re-laid. Also manually re-laying the progressView in `onLayout` didin't work as expected. A fix from [here](https://github.com/facebook/react-native/issues/17968) solved this issue.

### DIFF
--- a/packages/turbo/android/src/main/java/com/reactnativeturbowebview/RNVisitableView.kt
+++ b/packages/turbo/android/src/main/java/com/reactnativeturbowebview/RNVisitableView.kt
@@ -200,6 +200,10 @@ class RNVisitableView(context: Context, sessionModule: RNSessionModule) : Linear
     }
   }
 
+  /*
+   * Fixes a bug in React Native, causing improper layout of the TurboView and its children.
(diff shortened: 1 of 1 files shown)


packages/turbo/android/src/main/java/com/reactnativeturbowebview/RNVisitableView.kt
```diff
@@ -201,15 +200,26 @@ class RNVisitableView(context: Context, sessionModule: RNSessionModule) : Linear
...
+    post(measureAndLayout)
+  }
+
+  private val measureAndLayout = Runnable {
+    measure(
+      MeasureSpec.makeMeasureSpec(width, MeasureSpec.EXACTLY),
+      MeasureSpec.makeMeasureSpec(height, MeasureSpec.EXACTLY)
+    )
+    layout(left, top, right, bottom)
+  }
```
pfeiffer, line 214: Would it make sense to add a comment describing why this code has been added?

PR #53, opened 2023-11-05
Remove Session component

pklatka:
This PR is partially dependent on #43.

## Summary
This PR:
1. Merges Session component into `VisitableView`.  This will allow to use `VisitableView` as a standalone component. 
2. Modifies `react-native-web-screen` library to support setting custom `VisitableView`/`WebView` component for each route
3. Updates docs

### DIFF
--- a/packages/navigation/README.md
+++ b/packages/navigation/README.md
@@ -99,9 +99,29 @@ https://user-images.githubusercontent.com/25584895/225870138-b034f335-a30f-4e25-
 
 ## Advanced usage
 
-This library under the hood uses [react-native-turbo](../turbo/README.md). You can use React Navigation support (described here) or standalone React `VisitableView.tsx` component for more advanced cases.
+This library under the hood uses [react-native-turbo](../turbo/README.md). You can use React Navigation support (described here) or standalone React `VisitableView.tsx` component for more advanced cases. You can also define your own `WebScreen` component.
 
 Check out [react-native-turbo](../turbo/README.md) for more info.
+
+### Using custom WebScreen component
+
+You can use your custom `WebScreen` component by passing it to the `buildWebScreen` config. This can be useful if you want to define custom logic for each screen.
+
+```tsx
+import { WebView } from 'react-native-webview';
+
+const webScreens = buildWebScreen(webScreenConfig, {
+  webScreenComponent: WebScreen,
+});
+```
+
+To obtain `url` for current screen, use this code in the custom component:
+
+```tsx
+const path = route?.params?.path || route?.path;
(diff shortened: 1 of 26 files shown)


sebryu (approved):
I've put some comments into other PR https://github.com/software-mansion-labs/react-native-turbo-demo/pull/44

packages/navigation/src/buildWebScreen.tsx
```diff
@@ -10,16 +10,24 @@ export interface WebScreenRule {
...
 
-const buildWebviewComponent = (baseURL: string) => (navProps: any) =>
-  <WebScreen {...navProps} baseURL={baseURL} />;
+const buildWebviewComponent =
+  (baseURL: string, Component?: React.ElementType) => (navProps: any) =>
+    Component ? (
+      <Component {...navProps} baseURL={baseURL} />
+    ) : (
+      <WebScreen {...navProps} baseURL={baseURL} />
+    );
```
pfeiffer, line 30: What about just setting a default value for `Component` argument here?

packages/navigation/README.md
```diff
@@ -99,10 +99,30 @@ https://user-images.githubusercontent.com/25584895/225870138-b034f335-a30f-4e25-
...
+### Using custom WebScreen component
+
+You can use your custom `WebScreen` component by passing it to the `buildWebScreen` config. This can be useful if you want to define custom logic for each screen.
+
+```tsx
+import { WebView } from 'react-native-webview';
+
+const webScreens = buildWebScreen(webScreenConfig, {
+  webScreenComponent: WebScreen,
+});
```
pfeiffer, line 115: TBH I'm wondering if this is needed. I think the `webScreenComponent` prop doesn't really provide us any value, as we [define navigation Stack.Screens explicitly anyway](https://github.com/software-mansion-labs/react-native-turbo-demo/blob/fb1e755604a42ab04803c6684b0a663d5810d93e/example/src/App.tsx#L41-L82).

The prop here sets the `component` prop for all the screens as we hoist the screen props, which we might as well do 'manually', eg:

```jsx
const CustomVisitableView = (...) => <VisitableView ...>

export () => {
  return (
    <Stack.Navigator>
      <Stack.Screen
         ...
         component={CustomVisitableView} # We can simply supply it here instead of receiving it from the spread
      />
    </Stack.Navigator>
  )
}
```

pfeiffer: We could choose to address this later though. Might be good enough for now.

pklatka: Let's address this later.

PR #55, opened 2023-11-13
Support Strada

### LINKED ISSUE #41: Support for Strada
[Strada](https://strada.hotwired.dev/) was recently released, making it easier for the WebView pages to trigger native actions.

On the native side, "components" are registered and the web pages can communicate with those and trigger their actions. The native components can return data to the corresponding element on the web page.

The communication is built on top the message passing (JS-bridge) between the native side and WebView.

The registered components from the native side is also added to the user-agent string, so that the web site knows what components are available (see: https://github.com/hotwired/strada-ios/blob/d8c1bcf1d9511c03e7beed07c05d29f33f9b6cde/docs/QUICK-START.md?plain=1#L34-L35)

pklatka:
This PR depends on #53 and #44. Because of this reason, please review last three commits only!

## Summary
This PR resolves #41: it adds support for defining Strada components. Instead of adding `strada-android` and `strada-ios` to react-native-turbo, all JavaScript injections, event management, and component handling are performed on the React Native side.


packages/turbo/src/stradaBridge.ts
```diff
@@ -0,0 +1,223 @@
...
+  const stradaUserAgent = useMemo(
+    () =>
+      stradaComponents
+        ? ` bridge-components: [${stradaComponents.reduce(
+            (acc, { name }) => `${acc} ${name}`,
+            ''
+          )} ]`
+        : '',
+    [stradaComponents]
+  );
```
pfeiffer, line 163: This is really hard to read and reason about.

What about using `map` and `join` for the concatenation? Could also do an early return if no `stradaComponents` is nil.

pfeiffer:
```js
const stradaUserAgent = useMemo(() => {
  if (!stradaComponents) return;

  const componentNames = stradaComponents.map(({ name }) => name);
  return `bridge-components: [${componentNames.join(" ")}]`;
}, [stradaComponents]);

File: packages/turbo/src/stradaBridge.ts [hunk 2]
```diff
@@ -0,0 +1,223 @@
...
+
+    useEffect(() => {
+      const eventHandlerEventSubscription = registerMessageEventListener(
+        name,
+        didReceive
+      );
+      return () => {
+        eventHandlerEventSubscription.remove();
+      };
+    }, [name, onReceive, didReceive]);
```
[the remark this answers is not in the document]
pklatka: Wondering why `eslint` didn't figure out that this could be removed :)

packages/turbo/src/stradaBridge.ts
```diff
@@ -0,0 +1,223 @@
...
+        return;
+      }
+      Promise.all(
+        stradaComponents.map(({ name }) =>
+          RNVisitableViewModule.injectJavaScript(
+            sessionHandle,
+            `window.nativeBridge.register('${name}')`
+          )
+        )
+      );
```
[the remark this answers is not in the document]
pklatka: As far as I can see, the bridgeDelegate in both [strada-ios](https://github.com/search?q=repo%3Ahotwired%2Fstrada-ios%20unregister&type=code) and [strada-android](https://github.com/search?q=repo%3Ahotwired%2Fstrada-android%20unregister&type=code) does not use the unregister method anywhere.

pfeiffer:
I think we could consider implementing the `Form` and potentially the `Menu` + `Overflow + Menu` Strada components as in the example app. This would enable us to point the example app to the official Turbo Native server (https://turbo-native-demo.glitch.me) and remove the `example/server` directory 🎉

packages/turbo/src/stradaBridge.ts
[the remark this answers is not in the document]
pfeiffer: Hah, then let's leave it as is for now.

packages/turbo/README.md
```diff
@@ -76,6 +74,59 @@ The library implements a native view `RNVisitable` component for React Native. T
 
 URL for the WKWebview to open. Changing the url should result in view replacing opening different page.
 
+### `sessionHandle`
+
+Session handle for the webview. If not provided, the default session will be used. It can be used to create separate webview instances for different parts of the app.
+
+### `stradaComponents`
+
+`VisitableView` supports defining [`Strada components`](https://strada.hotwired.dev/) - components that receive and reply to messages from web components that are present on the page _within one session_. This prop accepts an array of `Strada` components that will be registered in the webview.
```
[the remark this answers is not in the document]
pklatka: Basically, if you were to write the app natively on Android or iOS, you would need to define a global array of Strada component factories to use Strada. The user-agent is later updated based on this global array.

Now, if the user wants to define two VisitableViews with the same `sessionHandle` (one native WebView instance) but different `stradaComponents`, the `userAgent` would update properly, however, this change is not detected by Strada on the web for some reason, which should be investigated. This might be connected to the different approach used when writing the app natively, where the userAgent is set once (when initialising the native WebView). That's why I've decided to highlight "within one session".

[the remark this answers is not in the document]
pfeiffer: I'm not sure we want to support having multiple instances of `VisitableView` with the same `sessionHandle`. I can't think of a reason why you'd want that.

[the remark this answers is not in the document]
pklatka: Yeah, you're right, I also can't think of the use case. I think it would make sense to remove the emphasis from that sentence.

packages/turbo/src/stradaBridge.ts
```diff
@@ -0,0 +1,223 @@
...
+              data: { ...previousMessage.data, ...data },
+            }
+          : {
+              component: name,
+              event,
+              data: {
+                metadata: {
+                  url,
+                },
+                ...data,
```
sebryu, line 204: i believe you should spread `data` before setting `metadata` (otherwise your metadata can be overridden by data.metadata)

packages/turbo/src/VisitableView.tsx
```diff
@@ -96,14 +107,34 @@ const VisitableView = React.forwardRef<RefObject, React.PropsWithRef<Props>>(
...
+      },
+      [initializeStradaBridge, onLoad]
+    );
+
+    const memoizedStradaComponents = useMemo(
+      () =>
+        stradaComponents?.map(({ StradaReactComponent }, i) => (
+          <StradaReactComponent key={i} useStrada={useStrada} />
+        )),
+      [stradaComponents, useStrada]
```
sebryu, line 123: currently `useStrada` returns function on each render, and we should probably wrap it in `useCallback`

example/src/Strada/Form.tsx
```diff
@@ -0,0 +1,44 @@
...
+import { BridgeComponent } from 'react-native-turbo';
+import { NavigationContext } from '@react-navigation/native';
+
+export default class Form extends BridgeComponent {
+  static name = 'form';
+  static contextType = NavigationContext;
+
+  constructor(props) {
+    super(props);
+  }
```
pfeiffer, line 12: This is not needed.

example/src/Strada/Form.tsx
```diff
@@ -0,0 +1,44 @@
...
+          onPress={() => this.replyTo('connect')}
+          title="Send"
+          disabled={disabled}
+        />
+      ),
+    });
+  }
+
+  onReceive(e) {
+    switch (e.event) {
```
pfeiffer, line 27: Consider changing `e` to `message` for clarity and be close to [Strada conventions](https://github.com/hotwired/strada-ios/blob/main/docs/BUILD-COMPONENTS.md?plain=1#L33-L34)

pfeiffer: Also consider defining and exporting the `StradaMessage` type

packages/turbo/README.md
```diff
@@ -78,6 +78,82 @@ URL for the WKWebview to open. Changing the url should result in view replacing
...
+The name of the application as used in the user agent string.
+
+### `stradaComponents`
+
+`VisitableView` supports defining [Strada components](https://strada.hotwired.dev/) that receive and reply to messages from web components that are present on the page within one session. This prop accepts an array of Strada components that will be registered in the webview.
+
+You can define a Strada component by:
+
+1. Extending `BridgeComponent` class and implementing `onReceive` method and static `name` property.
+2. Wrapping the component with `bridgeComponent` utility.
```
pfeiffer, line 92: I get that we have this, but I'm wondering if it should be officially exported by the library? It will need to be maintained going forward and I think the class definitions will be the preferred and recommended solution to keep closer to Strada native.

example/src/Strada/Form.tsx
```diff
@@ -0,0 +1,44 @@
...
+  constructor(props) {
+    super(props);
+  }
+
+  setHeaderButton(disabled: boolean) {
+    this.context.setOptions({
+      headerRight: () => (
+        <Button
+          onPress={() => this.replyTo('connect')}
+          title="Send"
```
pfeiffer, line 19: The Strada Web component provides a `submitTitle` as the message data, so let's use that in the form component as well:

```
// iOS
private func handleConnectEvent(message: Message) {
    guard let data: MessageData = message.data() else { return }
    configureBarButton(with: data.submitTitle)
}
```

packages/turbo/src/BridgeComponent.ts
```diff
@@ -0,0 +1,89 @@
...
+
+class BridgeComponent extends Component {
+  private name: string;
+  private url: string;
+  private sessionHandle: string;
+  private eventHandlerEventSubscription?: EmitterSubscription;
+  private previousMessages: StradaMessages = {};
+
+  constructor(props: StradaReactComponentProps) {
+    super(props);
```
pfeiffer, line 23: Nitpicking; add a line-break after the super call?

packages/turbo/src/BridgeComponent.ts
```diff
@@ -0,0 +1,89 @@
...
+);
+
+class BridgeComponent extends Component {
+  private name: string;
+  private url: string;
+  private sessionHandle: string;
+  private eventHandlerEventSubscription?: EmitterSubscription;
+  private previousMessages: StradaMessages = {};
+
+  constructor(props: StradaReactComponentProps) {
```
pfeiffer, line 22: Considering if we should throw an error if `name` is not defined. This is a required static for components to work.

packages/turbo/src/BridgeComponent.ts
```diff
@@ -0,0 +1,89 @@
...
+    this.onReceive(message);
+  }
+
+  onReceive(_e: StradaEvent) {
+    throw new Error(
+      `BridgeComponent.onReceive must be overridden by ${this.name}`
+    );
+  }
+
+  replyTo(event: string, data?: object) {
```
pfeiffer, line 58: Strada native provides a few helpers to reply to messages. We have implemented `replyTo` here, but we could consider adding the rest to have parity with Strada native: https://github.com/hotwired/strada-android/blob/main/docs/BUILD-COMPONENTS.md?plain=1#L103-L112

```kotlin
replyTo("eventName")
replyTo("eventName", newData)
replyWith(originalMessage)
replyWith(originalMessage.replacing(data = newData))
replyWith(originalMessage.replacing(event = "newEventName"))
```

packages/turbo/src/BridgeComponent.ts
```diff
@@ -0,0 +1,89 @@
...
+      this.name,
+      this.didReceive
+    );
+  }
+
+  componentWillUnmount() {
+    this.eventHandlerEventSubscription?.remove();
+  }
+
+  didReceive(e: object) {
```
pfeiffer, line 43: It's a tiny bit confusing to have both `didReceive` and `onReceive`.

I also think that it doesn't make sense to have `didReceive` to be defined for each BridgeComponent; I think it'd be better if the components only received messages that are relevant for their instance (ie. where the URL matches). The BridgeComponents should not have to care about ignoring messages meant for other instances than itself, it feels kinda leaky.

pfeiffer: Maybe add a helper, eg. `registerStradaMessageEventListener(component)` that sets up the event listener with URL filtering alá:

```js
function registerStradaMessageEventListener(component) {
   const listener = registerMessageEventListener(component.name, (e: object) => {
     const message = e as StradaEvent;
     if (message.data.metadata.url !== component.url) { return }
  
     component.onReceive(message)
   })
}

File: packages/turbo/src/BridgeComponent.ts [hunk 15]
```diff
@@ -0,0 +1,89 @@
...
+    this.url = props.url;
+    this.name = props.name;
+    this.sessionHandle = props.sessionHandle;
+    this.didReceive = this.didReceive.bind(this);
+    this.onReceive = this.onReceive.bind(this);
+    this.replyTo = this.replyTo.bind(this);
+  }
+
+  componentDidMount() {
+    this.eventHandlerEventSubscription = registerMessageEventListener(
```
pfeiffer, line 33: A bit funky name here (double `event`)

packages/turbo/src/BridgeComponent.ts
```diff
@@ -0,0 +1,89 @@
...
+  private sessionHandle: string;
+  private eventHandlerEventSubscription?: EmitterSubscription;
+  private previousMessages: StradaMessages = {};
+
+  constructor(props: StradaReactComponentProps) {
+    super(props);
+    this.url = props.url;
+    this.name = props.name;
+    this.sessionHandle = props.sessionHandle;
+    this.didReceive = this.didReceive.bind(this);
```
pfeiffer, line 27: Nitpicking, line break above this?

packages/turbo/src/BridgeComponent.ts
```diff
@@ -0,0 +1,89 @@
...
+    }
+    this.previousMessages[message.event] = message;
+    this.onReceive(message);
+  }
+
+  onReceive(_e: StradaEvent) {
+    throw new Error(
+      `BridgeComponent.onReceive must be overridden by ${this.name}`
+    );
+  }
```
pfeiffer, line 56: Do we want to throw an error if this is not overriden? I don't think so. It's not fatal if the method isn't defined in the subclass.

packages/turbo/README.md
pfeiffer: Removing the `bridgeComponent` would also allow us to clean up this: `const Component = stradaComponent.StradaComponent ?? stradaComponent;` etc.

Thinking about it I'm certain it would provide a clearer maintenance path to only support class components.

packages/turbo/README.md
```diff
@@ -78,6 +78,82 @@ URL for the WKWebview to open. Changing the url should result in view replacing
...
+```jsx
+import { BridgeComponent } from 'react-native-turbo';
+
+export default class Form extends BridgeComponent {
+  static name = 'form';
+
+  // Remember to call super(props) before registering this component
+  constructor(props) {
+    super(props);
+  }
```
pfeiffer, line 105: Not needed? We're not setting any default props or anything here anyway.

packages/turbo/README.md
```diff
@@ -78,6 +78,82 @@ URL for the WKWebview to open. Changing the url should result in view replacing
...
+
+  onReceive(message) {
```
pfeiffer, line 107: `message: StradaMessage`

packages/turbo/README.md
```diff
@@ -78,6 +78,82 @@ URL for the WKWebview to open. Changing the url should result in view replacing
...
+    // Here you can catch events from webview and respond to them
+    if (message.event === 'connect'){
```
pfeiffer, line 109: Missing space

packages/turbo/src/BridgeComponent.ts
pfeiffer: We can implement this later though if we're pressed on time :)

packages/turbo/src/BridgeComponent.ts
pklatka: `name` is always defined on a class (it defaults to the class name):

```
> class Test {}
> Test.name
> 'Test'
```

pfeiffer: Ah, awesome, TIL :-)

packages/turbo/src/types.ts
```diff
@@ -16,6 +16,16 @@ export interface VisitProposalError {
   error?: string;
 }
 
+export type StradaMessage = {
+  component: string;
+  event: string;
+  data: {
```
pfeiffer, line 22: I think `data` can contain more than just the `metadata` right?

example/src/Strada/Form.tsx
```diff
@@ -0,0 +1,42 @@
...
+      case 'submitDisabled':
+        this.setHeaderButton(true);
+        break;
+
+      case 'submitEnabled':
+        this.setHeaderButton(false);
+        break;
+
+      default:
+        break;
```
pfeiffer, line 39: Not needed?

packages/turbo/src/VisitableView.tsx
```diff
@@ -45,22 +48,30 @@ export interface RefObject {
...
+      stradaComponents,
       onVisitError: viewErrorHandler,
+      onLoad,
     } = props;
     const messageHandlerEventSubscription = useRef<EmitterSubscription>();
+    const { initializeStradaBridge, stradaUserAgent } = useStradaBridge(
+      sessionHandle,
+      stradaComponents
+    );
+    const resolvedApplicationNameForUserAgent = `${applicationNameForUserAgent} ${stradaUserAgent}`;
```
pfeiffer, line 64: I think we only want to append `stradaUserAgent` if it's non-empty.

PR #57, opened 2023-11-23
Refactor native code

pklatka:
## Summary

This PR: 
1. Removes RNSessionModule (RNVisitableViewModule)
2. Adds new onWarning prop to send warnings from native to JS side
3. Modifies the way we communicate from JS to native.

### DIFF
--- a/packages/turbo/src/VisitableView.tsx
+++ b/packages/turbo/src/VisitableView.tsx
@@ -3,6 +3,7 @@ import React, {
   useImperativeHandle,
   useRef,
   useMemo,
+  useEffect,
 } from 'react';
 import { NativeSyntheticEvent, StyleSheet } from 'react-native';
 import { getNativeComponent } from './common';
--- a/example/src/ShareScreen.tsx
+++ b/example/src/ShareScreen.tsx
@@ -3,13 +3,11 @@ import type { SessionMessageCallback } from 'react-native-turbo';
 import WebView, { type Props } from './WebView';
 import { Share } from 'react-native';
 
-type MessageType = { message: { method: 'share'; shareText: string } };
+type MessageType = { method: 'share'; shareText: string };
 
 const ShareScreen: React.FC<Props> = (props) => {
   const onMessage: SessionMessageCallback = async (message) => {
-    const {
-      message: { shareText },
-    } = message as MessageType;
+    const { shareText } = message as MessageType;
     Share.share({
       title: 'Share React Native Web Screen library',
       message: shareText,


packages/turbo/src/stradaBridge.ts
```diff
@@ -86,7 +84,7 @@ const stradaBridgeScript = `
       ${
         Platform.OS === 'android'
           ? 'StradaNative.postMessage(JSON.stringify(message))'
-          : 'webkit.messageHandlers.stradaNative.postMessage(message)'
+          : 'webkit.messageHandlers.nativeApp.postMessage(message)'
```
pklatka, line 87: Changing `userContentController` name to remove unnecessary `stradaNative` `userContentController` on native side.

packages/turbo/src/common.ts
```diff
@@ -18,16 +11,23 @@ export enum SessionEvents {
...
+  return {
+    [componentName]: requireNativeComponent<T>(componentName),
+    dispatchCommand: (ref, command, ...args) => {
+      UIManager.dispatchViewManagerCommand(
+        findNodeHandle(ref.current),
+        viewConfig.Commands[command],
+        args
+      );
+    },
+  };
```
pklatka, line 30: New way of native -> js communication. We dispatch an event using UIManager and catch it in the native view manager.

packages/turbo/ios/RNSessionManager.swift
```diff
@@ -0,0 +1,27 @@
...
+
+class RNSessionManager {
+
+  private var sessions: [NSString: RNSession] = [:]
+  static var shared: RNSessionManager = RNSessionManager()
+
+  private init() {}
+
+  func findOrCreateSession(sessionHandle: NSString?, webViewConfiguration: WKWebViewConfiguration) -> RNSession {
+    let resolvedSessionHandle = sessionHandle ?? ("Default" as NSString)
```
pfeiffer, line 18: I think we should make `sessionHandle` required and remove this. Why would we call `findOrCreateSession(..)` without an explicit handle?

JS Component could provide it's default handle, so having stricter requirements on native side I think is better.

packages/turbo/src/VisitableView.tsx
```diff
@@ -125,21 +103,32 @@ const VisitableView = React.forwardRef<RefObject, React.PropsWithRef<Props>>(
...
       <>
         {stradaComponents?.map((Component, i) => (
           <Component
             key={i}
+            url={url}
             sessionHandle={sessionHandle}
             name={Component.componentName}
-            url={url}
+            registerMessageListener={registerMessageListener}
+            sendToBridge={sendToBridge}
```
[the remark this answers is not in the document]
pfeiffer: I'm thinking this could be improved.

As discussed, I am hugely in favor of exposing `VisitableView#injectJavaScript` to run some JS in the WebView, and an `onMessage` prop which will be called whenever there's a message _from_ the WebView.

The reason is that this maps 1:1 with the `react-native-webview` props and how other RN components would typically do this (`onSomething`) instead of an array of listeners. `injectJavaScript` maps to `WKWebView` method as well.

packages/turbo/ios/RNSessionManager.swift
pklatka: > JS Component could provide it's default handle, so having stricter requirements on native side I think is better.
It actually does, will remove it.

packages/turbo/src/VisitableView.tsx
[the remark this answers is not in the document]
pfeiffer: A Strada HOC would simply provide an `onMessage` method, which would call the original if the message doesn't look like a Strada message:

```jsx
const withStrada =(Component) => ({ stradaComponents, onMessage, ...props }) => {
  const onMessageWithStrada = (message) => {
     // Do it's stuff
     
    onMessage?(message) // bubble to original prop
  }  

  return (
    <>
      .. strada components here
      <Component onMessage={onMessageWithStrada} {...props} />
    </>
  )
}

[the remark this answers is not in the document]
Comment (pklatka) [reply packages/turbo/src/VisitableView.tsx:120]: > As discussed, I am hugely in favor of exposing VisitableView#injectJavaScript to run some JS in the WebView, and an onMessage prop which will be called whenever there's a message from the WebView.

All messages from native WebView are being passed to `onMessage` prop on `VisitableView`.
We expose VisitableView#injectJavaScript in its `ref`:
```jsx
    useImperativeHandle(
      ref,
      () => ({
        injectJavaScript: (callbackStringified) => {
          dispatchCommand(
            visitableViewRef,
            'injectJavaScript',
            callbackStringified
          );
        },
      }),
      []
    );
```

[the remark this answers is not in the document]
pfeiffer: Awesome! What do we need `registerMessageListener` and `sendToBridge` in that case? Wouldn't `onMessage` and `injectJavaScript` be enough for us?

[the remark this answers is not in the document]
pfeiffer: Apologies, I see that the `registerMessageListener` and `sendToBridge` are on the `BridgeComponent`, not the `VisitableView` :-)

packages/turbo/ios/RNVisitableView.swift
```diff
@@ -10,22 +10,53 @@ import UIKit
...
+      sendSessionConfigurationWarning(propertyName: "webViewConfiguration")
+    }
+    return session
+  }()
+  private lazy var turboSession: Session = session.turboSession
+  private lazy var webView: WKWebView = turboSession.webView
+    
+  lazy var controller: RNVisitableViewController = {
+    let url = URL(string: String(url))!
+    let controller = RNVisitableViewController(url: url)
```
pfeiffer, line 52: Wondering if we could simplify this code by _not_ passing the URL here to the init'ed controller.

Instead, the URL of the controller, would be controller by a single place (`url` setter) and we might be able to drop the conditional `if (oldValue != "") ... ` in line 25.

packages/turbo/ios/RNVisitableView.swift
```diff
@@ -34,72 +65,64 @@ class RNVisitableView: UIView, SessionSubscriber {
...
-    guard let sessionModule = self.bridge?.uiManager.moduleRegistry.module(forName: "RNVisitableViewModule") as? RNVisitableViewModule else {
-        print("Couldn't find session for sessionHandle:", sessionHandle ?? "default session")
-        return nil
-    }
-    return sessionModule.getSession(sessionHandle: sessionHandle)
-  }
-  
-  public func getSession() -> Session? {
-    return getRNSesssion()?.turboSession
+  public func getWebViewConfiguration() -> WKWebViewConfiguration{
```
pfeiffer, line 69: We've changed most of the other accessor methods to `lazy var` - is there a reason why we haven't done this here?

packages/turbo/ios/RNVisitableView.swift
```diff
@@ -10,22 +10,53 @@ import UIKit
...
+      validateSessionProperty(propertyName: "applicationNameForUserAgent", oldValue: oldValue, newValue: applicationNameForUserAgent)
+    }
+  }
   @objc var url: NSString = "" {
     didSet {
       if (oldValue != "") {
-        self.controller?.visitableURL = URL(string: String(url))!
-        getSession()?.visit(controller!)
+        controller.visitableURL = URL(string: String(url))!
+        turboSession.visit(controller)
```
pfeiffer, line 27: I think this should be extracted into a method, eg. `visit(url)`

packages/turbo/ios/RNVisitableView.swift
```diff
@@ -34,72 +65,64 @@ class RNVisitableView: UIView, SessionSubscriber {
...
+  }
+    
+  private func sendSessionConfigurationWarning(propertyName: NSString){
+    onWarning?(["message": "You cannot change \(propertyName) after initialization of the session."])
+  }
+  
+  private func validateSessionProperty(propertyName: NSString, oldValue: NSString?, newValue: NSString?) -> Void{
+    if (oldValue != nil && newValue != nil  && !oldValue!.isEqual(to: newValue! as String)){
+      sendSessionConfigurationWarning(propertyName: propertyName)
+    }
```
pfeiffer, line 109: I'm inclined to say that this is a premature optimization having to maintain these (and also the corresponding `Session#configurationEquals` method. See also the comment in the `lazy var session` block.

packages/turbo/ios/RNSession.swift
```diff
@@ -11,61 +11,58 @@ import WebKit
...
       $0.id == newView.id
     }) {
-      registeredVisitableViews.append(newView)
+      visitableViews.append(newView)
     }
-    newView.attachDelegateAndVisit(newView.controller!)
+    newView.attachDelegateAndVisit(newView.controller)
   }
   
   func removeVisitableView(view: SessionSubscriber) {
```
pfeiffer, line 37: I find this method really hard to read and reason about. Maybe we could refactor it for better readability?

packages/turbo/ios/RNSessionManager.swift
```diff
@@ -0,0 +1,24 @@
...
+
+class RNSessionManager {
+
+  private var sessions: [NSString: RNSession] = [:]
+  static var shared: RNSessionManager = RNSessionManager()
+
+  private init() {}
+
+  func findOrCreateSession(sessionHandle: NSString, webViewConfiguration: WKWebViewConfiguration) -> RNSession {
+    if(sessions[sessionHandle] == nil){
```
pfeiffer, line 18: Missing space after `)`

packages/turbo/ios/RNVisitableView.swift
```diff
@@ -10,22 +10,53 @@ import UIKit
...
+  @objc var sessionHandle: NSString? = nil {
+    didSet {
+      validateSessionProperty(propertyName: "sessionHandle", oldValue: oldValue, newValue: sessionHandle)
+    }
+  }
+  @objc var applicationNameForUserAgent: NSString? = nil {
+    didSet {
+      validateSessionProperty(propertyName: "applicationNameForUserAgent", oldValue: oldValue, newValue: applicationNameForUserAgent)
+    }
+  }
```
pfeiffer, line 22: I think the property validation here are a bit too clever for now. I don't imagine that consuming developers would change these values dynamically and we add complexity here for very little gain.

packages/turbo/ios/RNVisitableView.swift
```diff
@@ -10,22 +10,53 @@ import UIKit
...
   @objc var url: NSString = "" {
     didSet {
       if (oldValue != "") {
```
pfeiffer, line 25: Might be able to drop this conditional, see https://github.com/software-mansion-labs/react-native-turbo-demo/pull/57/files#r1405894436

packages/turbo/ios/RNVisitableView.swift
```diff
@@ -34,72 +65,64 @@ class RNVisitableView: UIView, SessionSubscriber {
...
   
   override func didMoveToWindow() {
-
-    let url = URL(string: String(url))!
-    
-    if (self.controller == nil) {
-      print("Open new VistableView with URL passed from JS: ", url)
-      setupViewController(url: url)
-      self.addSubview((controller?.view)!)
+    if (!subviews.contains(controller.view)){
```
pfeiffer, line 76: Nitpicking, but is this conditional really needed? I suppose a single view can only be in the hierarchy once anyway.

packages/turbo/ios/RNVisitableView.swift
pklatka: 1. `VisitableViewController` doesn't have a constructor with no parameters. 
2. We could just pass an URL with empty string, but when you initialize `URL` object with empty string we obtain a `nil`, and  `VisitableViewController` force-unwraps the URL parameter and it would cause the app to crash.
3. If we would like to refactor the current `lazy` initialization to a method with url parameter, we would need to change the type of the `controller` variable to Optional (Swift need all properties to be initialized with a value) and add `?` in multiple places.
4. I don't think `if (oldValue != "")` check is hacky, it is necessary when you want to change url by changing `url` prop in RN. We also use it when you get the `replace` action from Turbo. 
5. We could extract `URL(...)` to a separate property, e.g. `visitableURL`, but this variable would be an Optional (reason in 2.). It would change the implementation to this:

```swift
  private var visitableURL: URL? = nil
    didSet {
      visitableURL = URL(string: url as String)
      if(visitableURL != nil){
        controller.visitableURL = visitableURL
      }
      if (oldValue != "") {
        turboSession.visit(controller)
      }
    }
  }

  lazy var controller: RNVisitableViewController = {
    let controller = RNVisitableViewController(url: visitableURL!)
    controller.delegate = self
    return controller
  }()  

  override func didMoveToWindow() {
    reactViewController().addChild(controller)
    controller.view.frame = bounds // Fixes incorrect size of the webview
    controller.didMove(toParent: reactViewController())
    addSubview(controller.view)
  }
    
```

packages/turbo/ios/RNVisitableView.swift
pklatka: No reason, we can use `lazy` here

packages/turbo/ios/RNVisitableView.swift
pklatka: If we would go back, we would replace an old instance with a new one, but this shouldn't matter to much.

packages/turbo/ios/RNVisitableView.swift
```diff
@@ -10,22 +10,53 @@ import UIKit
...
   
-  var controller: RNVisitableViewController?
+  var bridge: RCTBridge?
   
+  private lazy var session: RNSession = {
+    let webViewConfiguration = self.getWebViewConfiguration()
+    let session = RNSessionManager.shared.findOrCreateSession(sessionHandle: self.sessionHandle!, webViewConfiguration: webViewConfiguration)
+    if(session.configurationEquals(webViewConfiguration: webViewConfiguration)){
+      sendSessionConfigurationWarning(propertyName: "webViewConfiguration")
+    }
```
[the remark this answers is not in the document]
pfeiffer: I think keeping it as simple as possible here and add a warning to the docs is the way to go. Once things are stable and production-ready, we can always add these "nicenesses".

packages/turbo/ios/RNVisitableView.swift
```diff
@@ -8,98 +8,88 @@
...
       return
     }
-    session?.visitableViewWillAppear(visitable)
+    turboSession.visitableViewWillAppear(controller)
+  }
+
+  private func performVisitOnUrlChange(oldUrl: NSString){
+    if(oldUrl != "" && url != "" && oldUrl != url){
+      turboSession.visit(controller, action: .replace)
+    }
```
pklatka, line 76: Moved _initial load_ condition to `performVisit` method. I've tried to somehow avoid it but as said in the comment, session.visit function is called to set the currentVisit private variable.

pfeiffer: I think this is worse than before, haha. Now we have extracted this into a single-use method, that needs a dependent argument (`oldUrl`) to perform it's this.

packages/turbo/ios/RNVisitableView.swift
```diff
@@ -116,7 +106,7 @@ extension RNVisitableView: SessionDelegate {
     // Handle a visit proposal
     if (session.webView.url == proposal.url) {
```
pfeiffer, line 107: `webView.url` just?

packages/turbo/ios/RNVisitableView.swift
```diff
@@ -8,98 +8,88 @@
...
+  @objc var sessionHandle: NSString? = nil
+  @objc var applicationNameForUserAgent: NSString? = nil
   @objc var url: NSString = "" {
     didSet {
-      if (oldValue != "") {
-        self.controller?.visitableURL = URL(string: String(url))!
-        getSession()?.visit(controller!)
-      }
+      controller.visitableURL = URL(string: String(url))
+      performVisitOnUrlChange(oldUrl: oldValue)
```
pklatka, line 18: I know that we wanted to avoid checking the old value, however there is a problem with `replace` turbo action. On the RN side when `replace` action arrives, we call `setParams({url})` on react-navigation screen. Then this change is propagated to the `RNVisitableView` and no other Turbo session callback is being called. I was thinking that we could change this behavior, but we shoud maintain this communication order that we have today (proposal -> RN executes callback and propagates this to useWebviewNavigateHook -> do the visit on a native side). We could just add a line of code to the native onVisitProposal and check if action is equal to `replace` and call `turboSession.visit(...)` but we get two drawbacks:
1. We would no longer support defining custom behavior on RN side for `replace` event
2. We need to call `controller.visitableURL = URL(string: String(proposal.url))` before visit in the native onVisitProposal and we  also wanted to set it in `url didSet` only.

packages/turbo/ios/RNVisitableView.swift
pklatka: Reason [here](https://github.com/software-mansion-labs/react-native-turbo-demo/pull/57#discussion_r1407767067)

packages/turbo/ios/RNVisitableView.swift
pfeiffer: Actually, we get the `webView` from the passed-in `session` argument. While in theory, they point to the same, I guess it's cleaner to use the dependency from the argument as now.

packages/turbo/ios/RNVisitableView.swift
```diff
@@ -61,19 +60,11 @@ class RNVisitableView: UIView, RNSessionSubscriber {
...
     }
-    turboSession.visitableViewWillAppear(controller)
-  }
-
-  private func performVisitOnUrlChange(oldUrl: NSString){
-    if(oldUrl != "" && url != "" && oldUrl != url){
-      turboSession.visit(controller, action: .replace)
-    }
+    controller.visitableURL = URL(string: String(url))
+    turboSession.visit(controller)
```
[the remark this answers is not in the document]
pklatka: One issue that I spotted: closing the modal causes the scroll of previous screen from Stack navigator to stop working :/. Probably can be fixed by applying separate sessionHandle for the modal.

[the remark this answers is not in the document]
pklatka: Seems like an issue with ActivityIndicator - adding `controller.hideVisitableActivityIndicator()` after `turboSession.reload()` seems to fix the issue...

packages/turbo/ios/RNVisitableView.swift
```diff
@@ -8,98 +8,86 @@
...
-    // Upon initial load, the session.visit function is called
-    // to set the currentVisit private variable.
-    if(session?.webView.url == nil){
-      session?.visit(visitable)
-      return
-    }
-    session?.visitableViewWillAppear(visitable)
+    
+  public func performRestorationVisit(){
+    turboSession.visit(controller, action: .restore)
```
pklatka, line 74: Have to call .restore, otherwise the previous screen scroll position resets to the top of the page.

pfeiffer: I'm wondering if this would be solved by calling `session.visitableViewWillAppear` for the new top-most view after dismissing a modal?

It looks like `Session` has the exact behavior:

https://github.com/hotwired/turbo-ios/blob/c9d060fcdec136c3808cd0764e956d375012c6fb/Source/Session/Session.swift#L225-L226

Probably requires that `topmostVisit` in `Session` would point to the modal's `Visit`, which I assume is what we'd want anyway.

packages/turbo/ios/RNSession.swift
```diff
@@ -11,61 +11,54 @@ import WebKit
...
     }
-    newView.attachDelegateAndVisit(newView.controller!)
+    newView.becameTopMostView()
   }
   
-  func removeVisitableView(view: SessionSubscriber) {
-    // New view is not registered when presentation is modal
-    let isViewModal = registeredVisitableViews.last?.id == view.id
-    let viewIdx = registeredVisitableViews.lastIndex(where: {
+  func removeVisitableView(view: RNSessionSubscriber) {
```
pfeiffer, line 37: Suggestion; rename to `unregisterVisitableView`

packages/turbo/ios/RNSession.swift
```diff
@@ -48,7 +48,7 @@ class RNSession: NSObject {
         return
       }
       newView.becameTopMostView()
-      newView.performRestorationVisit()
+      newView.viewWillAppear()
```
pfeiffer: I'm sorry for keep digging in this, but we could still improve this. Why is `registerVisitableView` responsible for calling `becameTopMostView`? I think there's too much indirection here.

What about:

```swift
// RNVisitableView
func didBecomeTopMostView(restored: Bool) {
  session.delegate = self
  
  if (restored) {
    turboSession.visitableViewWillAppear(controller)
  }
}

func visitableWillAppear()
  session.registerVisitableView(...)
}

func visitableViewDidAppear() {
  didBecomeTopMostView(restored: false)
}

func visitableWillAppear() {
  session.unregisterVisitableView(...)
}

// RNSession
func registerVisitableView(..) {
  // Only concerned with registering views, not calling back to view, so we'll remove the `didBecomeTopMostView()` - this is instead called from the `visitableViewDidAppear`
}

func unregisterVisitableView() {
  // .. unregister here..

  if (wasTopMostView) {
    newView.didBecomeTopMostView(restored: true)
  }
}
```

This removes the uneeded circular indirection and side-effect in `registerVisitableView` - now that method is only concerned with registering the view. Setting up the delegate etc. is handled in `RNVisitableView` itself.

pfeiffer:
This looks good to me!

packages/turbo/ios/RNVisitableView.swift
```diff
@@ -8,115 +8,67 @@
...
   var id: UUID = UUID()
+  @objc var sessionHandle: NSString? = nil
+  @objc var applicationNameForUserAgent: NSString? = nil
   @objc var url: NSString = "" {
     didSet {
-      if (oldValue != "") {
-        self.controller?.visitableURL = URL(string: String(url))!
-        getSession()?.visit(controller!)
-      }
+      visit()
```
pklatka, line 17: When #62 will be ready to merge, we should probably remove this from didSet. I have two suggestions where we can add it:
1. Just move this to visitableWillAppear
2. Delegate making visits to RNSession. We would treat RNSessions as a “another manager for VistiableViews”. RNVisitableView could be treated as a “bridge” between RN and native.

packages/turbo/ios/RNSessionSubscriber.swift
```diff
@@ -11,7 +11,7 @@ protocol RNSessionSubscriber {
   
   var id: UUID { get set }
   var controller: RNVisitableViewController { get }
-  func didBecomeTopMostView(restored: Bool)
   func handleMessage(message: WKScriptMessage)
-  
+  func didProposeVisit(proposal: VisitProposal)
+  func didVisitFailed(visitable: Visitable, error: Error)
```
pfeiffer, line 16: This method name reads really weird.

packages/turbo/ios/RNSession.swift
```diff
@@ -11,61 +11,90 @@ import WebKit
...
+      guard let newView = visitableViews.last else {
+        return
+      }
+      visitableViewWillAppear(view: newView)
     }
   }
+    
+  func visitableViewWillAppear(view: RNSessionSubscriber) {
+    turboSession.visitableViewWillAppear(view.controller)
+  }
```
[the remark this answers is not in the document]
pfeiffer: I'm wondering if it makes sense to have a method here, that is only used once. I don't think `visitableViewWillAppear` would be called anywhere else than from `unregisterVisitableView`?

One other thing to consider is if we need this at all - if modals are supposed to have different session handles, the scenario where we'd need to re-register that the view behind a closing modal becomes the active one, isn't relevant.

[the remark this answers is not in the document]
pklatka: >I'm wondering if it makes sense to have a method here, that is only used once. I don't think visitableViewWillAppear would be called anywhere else than from unregisterVisitableView?

I've decided to wrap this into the separate method, because I think it is more readable when you call `visitableViewWillAppear(view: newView)` rather than calling a method directly on `turboSession` and passing a `controller` property of the new view.

> One other thing to consider is if we need this at all - if modals are supposed to have different session handles, the scenario where we'd need to re-register that the view behind a closing modal becomes the active one, isn't relevant.

Hmm, no strong opinion here, we can either remove it or leave it.

packages/turbo/src/BridgeComponent.ts
```diff
@@ -1,48 +1,44 @@
...
-const registerStradaMessageEventListener = (component: BridgeComponent) =>
-  registerMessageEventListener(component.name, (e: object) => {
-    const message = e as StradaMessage;
-    if (message.data.metadata.url !== component.url) {
-      return;
-    }
-    component.previousMessages[message.event] = message;
-    component.onReceive(message);
-  });
+const stradaMessageListener = (component: BridgeComponent) => (e: object) => {
```
sebryu, line 9: any reasons why this isn't class method inside `BridgeComponent` ? then you would have component available as `this`

packages/turbo/src/stradaBridge.ts
```diff
@@ -119,32 +117,24 @@ const stradaBridgeScript = `
 `;
 
 const useStradaBridge = (
-  sessionHandle: string,
+  visitableViewRef: React.RefObject<any>,
+  dispatchCommand: DispatchCommand,
   stradaComponents?: StradaComponent[]
 ) => {
   const initializeStradaBridge = useCallback(async () => {
```
sebryu, line 124: you no longer need `async` here

packages/turbo/src/BridgeComponent.ts
```diff
@@ -1,48 +1,44 @@
...
+    this.sendToBridge = props.sendToBridge;
 
     this.onReceive = this.onReceive.bind(this);
     this.replyTo = this.replyTo.bind(this);
   }
 
   componentDidMount() {
-    this.messageEventListenerSubscription =
-      registerStradaMessageEventListener(this);
+    this.registerMessageListener(stradaMessageListener(this));
```
sebryu, line 41: I have a problem with having `register` with no `unregister` on unmount

sebryu (approved):
great progress Patryk 👏  See my suggestions - probably they are not blocking and could be added in consecutive PR

packages/turbo/src/BridgeComponent.ts
pklatka: https://github.com/software-mansion-labs/react-native-turbo-demo/pull/55#discussion_r1397029729

packages/turbo/ios/RNVisitableView.swift
```diff
@@ -8,115 +8,67 @@
...
-    getRNSesssion()?.removeVisitableView(view: self)
+  public func injectJavaScript(code: NSString) -> Void {
+    webView.evaluateJavaScript(code as String)
   }
-
-  func visitableDidRender(visitable: Visitable) {
-    guard let session = getSession() else {
+    
+  private func visit() {
+    if(controller.visitableURL?.absoluteString == url as String) {
```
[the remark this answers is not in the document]
pklatka: It is handled in `didProposeVisit`

PR #61, opened 2023-12-01
Create navigation stack dynamically

mironiasty:


example/src/App.tsx
```diff
@@ -77,6 +62,7 @@ const App: React.FC = () => {
           component={NestedTab}
           options={{ title: 'Nested Top Tab' }}
         />
+        {webStackScreen(Stack, webScreenConfig, Routes.Flow)}
```
mironiasty, line 65:
```suggestion
        {webScrens(Stack, webScreenConfig, Routes.Flow)}
```

pfeiffer: Maybe this could be mixed with `Stack.Group`? Also, wondering if we need the 

```jsx
<Stack.Group screenOptions={{ presentation: "modal" }}>
  {webScreens(Stack, webScreenConfig[Routes.Flow])}
</Stack.Group>
```

mironiasty: `webScreens` is using internally different navigation structure for that case. As I'm thinking about it, I think I would go with sth like this:
```ts
{webScreens(Stack, Routes.Flow, { type: 'modalFlow' )}
````

PR #63, opened 2023-12-04
Extract RNVisitableView and RNVisitableViewModule to separate files

mironiasty:
Refactor how `RNVisitableView` and `RNVisitableViewModule` are created. This should reduce number of app crashes  when package files are edited during development (with Hot Reload enabled), due to error:
```
Invariant Violation: Tried to register two views with the same name RNVisitableView, js engine: hermes
```

I've also added proper types to `RNVisitableView`, so they map what is passed to native code.

### DIFF
--- a/packages/turbo/src/BridgeComponent.ts
+++ b/packages/turbo/src/BridgeComponent.ts
@@ -1,16 +1,12 @@
 import { Component } from 'react';
 import type { EmitterSubscription } from 'react-native';
 import type {
-  VisitableViewModule,
   StradaMessage,
   StradaMessages,
   StradaComponentProps,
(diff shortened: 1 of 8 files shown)


sebryu:
@mironiasty RNVisitableView is being refactored in Partyk's PR - https://github.com/software-mansion-labs/react-native-turbo-demo/pull/57/files did you check if the issue still occurs on that branch?

mironiasty:
I've missed JS part of that refactor. Thanks for pointing that out. I think that issue will still be there, as PR has same way of creating components. I think I will wait for that PR to be completed and I will fix it on refactored code.

PR #66, opened 2023-12-05
New API for defining screens

mironiasty:


packages/navigation/src/types.ts
```diff
@@ -0,0 +1,21 @@
+import type { NativeStackNavigationOptions } from '@react-navigation/native-stack';
+
+export interface WebScreenRule {
+  urlPattern: string;
+  title?: string;
+  presentation?: NativeStackNavigationOptions['presentation'];
+  component?: React.ElementType;
+  options?: NativeStackNavigationOptions;
```
pfeiffer, line 8: Is `title` and `presentation` needed here as separate keys? What about just providing them part of the `options`?

packages/navigation/src/types.ts
```diff
@@ -0,0 +1,21 @@
+import type { NativeStackNavigationOptions } from '@react-navigation/native-stack';
+
+export interface WebScreenRule {
+  urlPattern: string;
+  title?: string;
+  presentation?: NativeStackNavigationOptions['presentation'];
+  component?: React.ElementType;
```
pfeiffer, line 7: I'm wondering if it should be the responsibility of the Web Screens to build the linking/routes for the paths with native screens (ie. a non-`VisitableView` component of the `Screen`) or instead that should be 'normal' React Navigation.

pfeiffer: A use case for the `component` option is to eg. a native screen on certain rourtes. We'd probably want the path/urlPattern to match a dynamic param, eg:

```js
{
   [Routes.Profile]: {
      urlPattern: "users/:id",
      component: ProfileScreen
   }
}
```

In this case, navigating to `/profile/mary` would render the `ProfileScreen` and we'd have `baseURL`, `path` *and* `id` available as params? I guess this could be as expected and working.

pfeiffer: If we want Web Screens to take ownership of generating the linking configuration for the whole NavigationContainer, including for routes which are native and not rendered as a webview, we should probably allow the options to match the linking options for React Navigation, ie. `parse`, `exact`, .. (`PathConfig`).

Alternatively we should think of a pattern where the Web Screen can be smoother 'mixed in' between a regular linking configuration and screen definition.

packages/navigation/src/types.ts
mironiasty: I left it as it was. But after refactor it indeed seems better solutions (to move it to options)

packages/navigation/src/useWebScreen.tsx
```diff
@@ -0,0 +1,94 @@
...
+        {Object.entries(filteredRoutes).map(([routeName, rule]) => {
+          if (isRule(rule)) {
+            const { urlPattern, component, options } = rule;
+            return (
+              <Navigator.Screen
+                key={routeName}
+                name={routeName}
+                //  @ts-expect-error Use proper typing for main components
+                component={component ?? defaultComponent}
+                initialParams={{ baseURL, path: urlPattern }}
```
pfeiffer, line 68: I think we should consider how we pass in the URL to visit to the screens here. As we don't update `params` doing `replace` actions anymore, it could be that we'd be able to structure it better.

In `WebScreen` we have these lines:

```js
const path = route?.params?.path || route?.path || '';
const baseURL = route?.params?.baseURL || props?.baseURL;
```

.. which I'm not a particular fan of and would like to see refactored.

I'm wondering if it'd be possible to get the _current navigation pathname_ as a hook, similar to Expo Router's `usePathname()`?

```jsx
function WebScreen() {
  const pathname = usePathname();
  ...
}
```

pfeiffer: The `urlPattern` name is kinda confusing here as well as we discussed on the call. It's not _really_ a pattern (it's a special-syntax path) and it's not a url (it's a part of a path) **and** we pass it onto the screen as `params.path`

mironiasty:
After additional tests on this approach - I am afraid that it will not work :(

PR #67, opened 2023-12-05
Disable link preview on iOS

mironiasty:
When user press longer on any link in app, Safari shows link preview by default. Let's disable it in app for now (in future we might want enable it as feature depending on configuration?)

| Before | After |
|--------|--------|
|  |  |


pfeiffer:
LGTM!

pklatka (approved):

PR #70, opened 2023-12-06
Unify VisitableView props and extract logic to hooks

mironiasty:
There are two important changes in this PR:

### Update API to `VisitableView` component
`onVisitProposal` and `onLoad` props will no longer required to have nativeEvent params. Instead they will accept just payload param (same as :`onVisitError` and `onLoad` did)

### Create `useCustomUrl` hook

This hook will replace manual getting current url from navigation route.

#### Refactors
I extracted hooks from VisitableView into separate files in `hooks` directory. Also navigation package will have their own `hooks` directory with existing hooks.

### DIFF
--- a/packages/turbo/src/types.ts
+++ b/packages/turbo/src/types.ts
@@ -10,9 +10,7 @@ export interface LoadEvent {
   url: string;
 }
 
-export interface MessageEvent {
-  message: object;
-}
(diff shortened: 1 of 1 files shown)


pklatka (approved):
LGTM! Before merging just change the `MessageEvent` format.

PR #71, opened 2023-12-06
Refactor native code on Android

pklatka:
## Summary
This PR is a follow-up to #57: refactors the Android native code and simplifies the message object format (no longer need to wrap the object with `{message: ...}`)


packages/turbo/ios/RNVisitableView.swift
```diff
@@ -95,6 +91,7 @@ extension RNVisitableView: RNVisitableViewControllerDelegate {
   
   func visitableWillAppear(visitable: Visitable) {
     session.registerVisitableView(newView: self)
+    visit()
```
pklatka, line 94: Moving visit() here to match Android code.

packages/turbo/android/src/main/java/com/reactnativeturbowebview/RNSession.kt
```diff
@@ -1,102 +1,156 @@
...
+    visitableViews.last().visitLocationStarted(location)
+  }
+
+  override fun visitProposedToLocation(location: String, options: TurboVisitOptions) {
+    visitableViews.last().visitProposedToLocation(location, options)
+  }
+
+  override fun visitRendered() {
+    visitableViews.last().visitRendered()
+  }
```
pfeiffer, line 152: Could we end up with `visitableViews.last()` returning `null` in the case that a VisitableView disappears and thus is unregistered before a visit is finished? In that case, would we call `visitCompleted()` on a null object?

packages/turbo/android/src/main/java/com/reactnativeturbowebview/RNSession.kt
```diff
@@ -1,102 +1,156 @@
...
-
-      for (view in registeredVisitableViews) {
-        view.detachWebView() {
-          onDetached()
-        }
-      }
-    }
-
-    if (!registeredVisitableViews.contains(newView)) {
-      registeredVisitableViews.add(newView)
```
pfeiffer, line 67: Wow, great to see how readable this becomes after the refactor 🎉

packages/turbo/android/src/main/java/com/reactnativeturbowebview/RNSession.kt
```diff
@@ -1,102 +1,156 @@
 package com.reactnativeturbowebview
```
pfeiffer, line 1: Should we consider renaming this to be more in-line with the package name(s)?

packages/turbo/src/VisitableView.tsx
```diff
@@ -105,14 +105,11 @@ const VisitableView = React.forwardRef<RefObject, React.PropsWithRef<Props>>(
...
-      (e: NativeSyntheticEvent<{ message: object }>) => {
-        onMessageCallbacks.current.forEach((listener) => {
-          listener?.(e.nativeEvent);
-        });
-      },
-      []
-    );
+    const handleOnMessage = useCallback((e: NativeSyntheticEvent<object>) => {
+      onMessageCallbacks.current.forEach((listener) => {
+        listener?.(e.nativeEvent);
```
mironiasty, line 110: note for myself: we can remove nullability for listeners to simplify code 🤔

packages/turbo/android/src/main/java/com/reactnativeturbowebview/RNSession.kt
pklatka: As discussed earlier, we will probably rename the whole library so let's leave it for now.

mironiasty (approved):

PR #72, opened 2023-12-07
Implement shared process pool

pklatka:
This PR sets a shared process pool on the `WKWebViewConfiguration`, which ensures that we share the same resources if multiple WKWebViews are in the app (ie. by multiple Sessions))


pklatka:
👀

PR #73, opened 2023-12-07
Simplify navigation definition

mironiasty:
Simplify how navigation is defined in projects using `react-native-web-screens`. Instead of relying on our custom object `WebScreenRuleConfig` and having to 'copy' it's definition to `react-navigation` structure - we can just rely on navigator definitions.
- we are removing custom `webScreens` object. Instead all navigation nodes are created manually with `<Nav.Screen .../>` definition
- app no longer rely on custom `WebScreenRuleConfig`, instead it uses same definition as [`linking`](https://reactnavigation.org/docs/navigation-container/#linking) object in `react-navigation`
- fixes transitions between screen (especially if `modal` is involved)


packages/navigation/src/hooks/useCurrentUrl.ts
```diff
@@ -1,16 +1,44 @@
...
 
   const currentRoute = state.routes[state.index];
-  const params = currentRoute?.params as
-    | { path?: string; baseURL?: string }
-    | undefined;
-
-  const path = params?.path ?? '';
-  const baseURL = params?.baseURL ?? defaultBaseUrl ?? '';
+  const path =
+    getPath(currentRoute?.params) ?? findPath(currentRoute?.name, config) ?? '';
```
pfeiffer, line 41: For fallback routes, path is available as `currentRoute.path` and not as `currentRoute.params.path`

pfeiffer, line 41:
```suggestion
  const path =
    getPath(currentRoute) ?? getPath(currentRoute?.params) ?? findPath(currentRoute?.name, config) ?? '';
```

I'm wondering if it could be made better. Hm.

packages/navigation/src/hooks/useCurrentUrl.ts
```diff
@@ -1,16 +1,44 @@
...
+}
+
+function getPath(params: unknown) {
+  if (params && typeof params === 'object' && 'path' in params) {
+    return params.path;
+  }
+  return undefined;
+}
+
+export function useCurrentUrl(baseUrl: string, config: LinkingConfig) {
```
pfeiffer, line 35: We must take params into consideration here, to support the use case where we navigate to a fallback route with params, eg: `/one_page?some_param=true`

In this case, the `getCurrentUrl()` should include the params as well and not just the path.

packages/navigation/src/hooks/useCurrentUrl.ts
```diff
@@ -1,16 +1,44 @@
...
   const state = navigation.getState();
 
   const currentRoute = state.routes[state.index];
-  const params = currentRoute?.params as
-    | { path?: string; baseURL?: string }
-    | undefined;
-
-  const path = params?.path ?? '';
-  const baseURL = params?.baseURL ?? defaultBaseUrl ?? '';
+  const path =
```
pfeiffer, line 40: I'm seeing some issues here with fallback screens that have multiple directories. What I'm seeing is that `currentRoute.params.path` is an array of paths, eg. if the fallback route matched is `/signup/form` then `params.path` is `["signup", "signup/form"]` which causes the final current URL to be `[...]/signup,signup/form`

pfeiffer:
Did some testing here today, and I think a solution for getting the navigated current URL from the `useCurrentUrl(), including query params etc. would be to associate it as a special param _before_ getting the state. I tried something like:

```js
// buildWebScreen.js
getLinkingObject() {
  return {
    config: ..
    getStateFromPath(path: string, options?: Options) {
      // Store the navigated path as an (encoded) query parameter so that we can
      // extract it later in getCurrentUrl()
      const webScreenParam = '__webScreen[path]=${encodeURIComponent(path)}'
      const pathWithWebScreenParam = `${path}${path.includes("?") ? "&" : "?"}${webScreenParam}`;

      return getStateFromPath(pathWithWebScreenParam, options);
    }
  }
}

// useCurrentUrl:
function getPath(params: unknown) {
  if (params && typeof params === 'object' && '__webScreen[path]') {
    return decodeURIComponent(params['__webScreen[path]']);
  }
  return undefined;
}
```

This makes both the params parsing by React Navigation work as expected for routes such as `/users/:slug` and also makes whatever query parameters are on the linked URL available as params, eg. `/users/jack?foo=boo` will have `{ slug: "jack", foo: "boo" }` as `params`.

packages/navigation/src/hooks/useCurrentUrl.ts
mironiasty: I think this should be fixed by now

packages/navigation/src/hooks/useWebviewNavigate.ts
```diff
@@ -68,47 +84,34 @@ export function useWebviewNavigate<
...
       }
 
       const { options } = linking;
 
       let path = to;
-      if (to.match(/^https?:\/\//)) {
-        path = extractPathFromURL(options?.prefixes, to) ?? '';
+      if (options?.prefixes && to.match(/^https?:\/\//)) {
+        path = extractPathFromURL(options.prefixes, to) ?? '';
       }
```
pfeiffer, line 98: Could be out-of-scope for this PR, but I find this confusing. We are stripping the host prefix, but only if the proposed URL contains `https://`? It feels like a wrong predicate here. I think it should more likely be `baseURL` that we checked for?

mironiasty: As this is pre-existing logic - let's fix it in separate PR.

pklatka (approved):
Great work!

PR #74, opened 2023-12-08
Introduce a reload method to the VisitableView 

pklatka:
This PR introduces a `reload` method to the `VisitableView`. This feature might be useful in scenarios where an error has occurred and reloading the page would be the desired recovery action.


mironiasty (approved):
👍

PR #75, opened 2023-12-09
Fix refresh and page restoration bugs on Android

pklatka:
This PR addresses two minor bugs that were occurring within the Android application:
1. The refresh view was not being displayed after navigating back to the previous screen.
2. There were issues with page restoration.

App behavior before fix:

https://github.com/software-mansion-labs/react-native-turbo-demo/assets/44744462/48f8617a-bf54-4dfb-a7f9-fd8d467feef1

After fix:

https://github.com/software-mansion-labs/react-native-turbo-demo/assets/44744462/ddcd2a0c-b0c6-411b-966a-333b01d9c5f5


mironiasty (approved):
I didn't test it, but change looks good 👍

PR #76, opened 2023-12-11
Add support for `alert` and `confirm`

mironiasty:
Add new props to VisitableView to support web `alert` and `confirm` dialogs.

There is also default support for both (that was missing before). So both `alert` and `confirm` will now work as React Native's `Alert.alert` call.


packages/turbo/src/VisitableView.tsx
```diff
@@ -107,11 +115,40 @@ const VisitableView = React.forwardRef<RefObject, React.PropsWithRef<Props>>(
...
+        const dispatch = (value: boolean) =>
+          dispatchCommand(
+            visitableViewRef,
+            'sendConfirmResult',
+            value.toString()
+          );
+        if (onConfirm) {
+          onConfirm(message, (value) => dispatch(value));
+        } else {
+          Alert.alert('Confirm dialog', message);
```
pfeiffer, line 145: We probably want the default behavior to show the message as the title and with two buttons; `OK` and `Cancel`

packages/turbo/src/hooks/useWebViewDialogs.ts
```diff
@@ -0,0 +1,52 @@
...
+  onConfirm: OnConfirm | undefined
+) {
+  const handleAlert = useCallback(
+    ({ nativeEvent: { message } }: NativeSyntheticEvent<AlertHandler>) => {
+      const dispatch = () =>
+        dispatchCommand(visitableViewRef, 'sendAlertResult');
+      if (onAlert) {
+        onAlert(message, () => dispatch());
+      } else {
+        Alert.alert(message, undefined, [{ text: 'Ok', onPress: dispatch }]);
```
pfeiffer, line 24:
```suggestion
        Alert.alert(message, undefined, [{ text: 'OK', onPress: dispatch }]);
```

packages/turbo/src/hooks/useWebViewDialogs.ts
```diff
@@ -0,0 +1,52 @@
...
+        dispatchCommand(
+          visitableViewRef,
+          'sendConfirmResult',
+          value.toString()
+        );
+      if (onConfirm) {
+        onConfirm(message, (value) => dispatch(value));
+      } else {
+        Alert.alert(message, undefined, [
+          { text: 'Ok', onPress: () => dispatch(true) },
```
pfeiffer, line 42:
```suggestion
          { text: 'OK', onPress: () => dispatch(true) },
```

pfeiffer (approved):

packages/turbo/ios/RNSession.swift
```diff
@@ -9,11 +9,32 @@ import RNTurboiOS
...
 import WebKit
 
+protocol WKUIDelegateHandler{
+  func alertHandler(message: String, completionHandler: @escaping () -> Void)
+  func confirmHandler(message: String, completionHandler: @escaping (Bool) -> Void)
+}
+
+class TurboUIDelegate: NSObject, WKUIDelegate{
+  private var uiHandler:WKUIDelegateHandler
+  init(uiHandler: WKUIDelegateHandler) {
```
pklatka, line 19:
```suggestion
class TurboUIDelegate: NSObject, WKUIDelegate {
  private var uiHandler:WKUIDelegateHandler
  
  init(uiHandler: WKUIDelegateHandler) {
```

packages/turbo/ios/RNSession.swift
```diff
@@ -9,11 +9,32 @@ import RNTurboiOS
...
+    self.uiHandler = uiHandler
+  }
+  
+  func webView(_ webView: WKWebView, runJavaScriptAlertPanelWithMessage message: String, initiatedByFrame frame: WKFrameInfo, completionHandler: @escaping () -> Void) {
+    self.uiHandler.alertHandler(message: message, completionHandler: completionHandler)
+  }
+  
+  func webView(_ webView: WKWebView, runJavaScriptConfirmPanelWithMessage message: String, initiatedByFrame frame: WKFrameInfo, completionHandler: @escaping (Bool) -> Void){
```
pklatka, line 27:
```suggestion
  func webView(_ webView: WKWebView, runJavaScriptConfirmPanelWithMessage message: String, initiatedByFrame frame: WKFrameInfo, completionHandler: @escaping (Bool) -> Void) {
```

pklatka (approved):

packages/turbo/ios/RNSession.swift
```diff
@@ -22,9 +43,11 @@ class RNSession: NSObject {
 
   public lazy var turboSession: Session = {
     webViewConfiguration.userContentController.add(self, name: "nativeApp")
+    self.wkUiDelegate = TurboUIDelegate(uiHandler: self)
     let session = Session(webViewConfiguration: webViewConfiguration)
     session.delegate = self
     session.webView.allowsLinkPreview = false
+    session.webView.uiDelegate = self.wkUiDelegate
```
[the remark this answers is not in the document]
mironiasty: I did it because uiDelegate is weak property. Is there better way of making it correct in swift? 🤔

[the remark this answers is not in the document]
pklatka: > I did it because uiDelegate is weak property.
You're right, let's leave it as it is.

PR #77, opened 2023-12-12
Enable WebView debugging

pklatka:
## Summary
This PR enables debugging the WebView when app is in a `DEBUG` mode.

### DIFF
--- a/packages/turbo/android/src/main/java/com/reactnativeturbowebview/RNSession.kt
+++ b/packages/turbo/android/src/main/java/com/reactnativeturbowebview/RNSession.kt
@@ -2,6 +2,7 @@ package com.reactnativeturbowebview
 
 import android.webkit.JavascriptInterface
 import android.webkit.WebSettings
+import android.webkit.WebView
 import androidx.appcompat.app.AppCompatActivity
 import androidx.lifecycle.Lifecycle
 import androidx.lifecycle.LifecycleOwner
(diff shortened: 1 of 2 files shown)


packages/turbo/src/VisitableView.tsx
```diff
@@ -129,6 +131,7 @@ const VisitableView = React.forwardRef<RefObject, React.PropsWithRef<Props>>(
           ref={visitableViewRef}
           url={props.url}
           sessionHandle={sessionHandle}
+          webviewDebuggingEnabled={webviewDebuggingEnabled}
```
mironiasty, line 134: We use `webView` in most other places. So maybe it would be better to have 
```suggestion
          webViewDebuggingEnabled={webViewDebuggingEnabled}
```

mironiasty (commented):
Maybe we can always enable debugging webview when app is in debugging mode (and remove `webviewDebuggingEnabled` from `VisitableView`).

This API makes it easy to set debugging enabled for production builds, which might not be desired option.

packages/turbo/src/VisitableView.tsx
pklatka: > Maybe we can always enable debugging webview when app is in debugging mode
> ...
> This API makes it easy to set debugging enabled for production builds, which might not be desired option.

It's definitely a better idea.

mironiasty (approved):
LGTM

PR #81, opened 2023-12-21
Fix app crash when using many tabs

mironiasty:
When app that uses `react-native-turbo` has many tabs, it often crashes. It is caused by calling `RNSession.unregisterVisitableView` when `visitableViews` list is empty.

@pklatka investigated it and found that `visitableWillAppear` is sometimes not called. So we can call it manually with this fix.

(related bugs: [one](https://stackoverflow.com/questions/131062/iphone-viewwillappear-not-firing) and [two](https://developer.apple.com/forums/thread/76834))

### DIFF
--- a/packages/turbo/ios/RNVisitableView.swift
+++ b/packages/turbo/ios/RNVisitableView.swift
@@ -47,6 +47,9 @@ class RNVisitableView: UIView, RNSessionSubscriber {
   }
   
   override func willMove(toWindow newWindow: UIWindow?) {
+    // visitableWillAppear is not called automatically sometimes. So it has to be called
+    // manually to make sure that visitableViews list is not empty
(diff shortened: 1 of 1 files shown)


packages/turbo/ios/RNVisitableView.swift
```diff
@@ -45,6 +45,10 @@ class RNVisitableView: UIView, RNSessionSubscriber {
     controller.didMove(toParent: reactViewController())
     addSubview(controller.view)
   }
+  
+  override func willMove(toWindow newWindow: UIWindow?) {
+    controller.viewWillAppear(false)
+  }
```
pfeiffer, line 51: Consider adding a comment here on why we have this, as it does look unconventional.

mironiasty: 👍

pklatka (approved):
🎉

PR #86, opened 2023-12-22
Do not call viewWillAppear directly on a view controller

pklatka:
## Summary
This PR modifies the approach used in #81. Directly invoking `viewWillAppear` on a view controller is not supported, and may lead to out-of-order callbacks and other inconsistent behaviors.


pklatka:
This PR also solves issues with scroll restorations introduced in #81, so I'm merging it now.

PR #89, opened 2023-12-24
Fix animations on `replace` actions

pfeiffer:
This PR changes how `replace` action is handled. Previously we always dispatched a replace navigation action with animations disabled (which doesn't _really_ work all the time). After this change, we only dispatch a replace navigation action in case the new route is different from the current route. If the route is the same, we can just update the params of the screen with the new replaced URL.

This fixes issues with animations and double-renders of the replaced screen.


pfeiffer:
This probably would require that the `RNVisitableView` supports fetching new page after updating the `url` prop.

pfeiffer:
Superseeded by #90

PR #90, opened 2024-01-04
 Fix animations on `replace` actions

mironiasty:
if REPLACE action will replace only URL for website, but not a screen in navigation - let's just replace URL without any 'real' navigation action.

This PR also adds missing logic to RNVisitableView - so when `url` prop is updated - it will open new url in WebView.


pfeiffer:
LGTM! We need to address how the `VisitableView` handles updates to the `url` prop in order to actually trigger a new visit when the prop is changed due to params being updated.

PR #91, opened 2024-01-04
Add form submission callbacks and snapshot clearing method

pklatka:
This PR:
1. Bumps `turbo-ios` version to `7.0.1` by publishing new fork to CocoaPods repository. Previous fork didn't have form submission callbacks implemented in `SessionDelegate`.
2. Exposes `didStartFormSubmission` and `didFinishFormSubmission` callbacks on `VisitableView`.
3. Adds `clearSnapshotCache` method to `VisitableView` which clears snapshots on a session which `VisitableView` operates.

### DIFF
--- a/packages/turbo/README.md
+++ b/packages/turbo/README.md
@@ -179,6 +179,20 @@ Note that after handling confirm dialog display, `callback` function must be cal
 - message
 - callback
 
+### `onFormSubmissionStarted`
+
+Callback called when website inside WebView started submitting form.
+
+- url
+
+### `onFormSubmissionFinished`
+
+Callback called when website inside WebView finished submitting form.
+
+- url
+
+Note: The form submission handlers are triggered for the _session_ in which the form was submitted. A URL argument is available in these handlers, which can be used for granular control over the cache clearing process. For example, you might choose to clear the cache only for the specific URL that the form was submitted from.
+
 ### Methods:
 
 ### `injectJavaScript(jsCode)`
--- a/packages/turbo/README.md
+++ b/packages/turbo/README.md
@@ -27,6 +27,8 @@ For Android you need to adjust your SDK version in your `build.gradle`.
 >
 > Android SDK 24+ is required as the minSdkVersion in your build.gradle.
 
+For iOS, you need to set the deployment target to 14.0 or higher.
+
 ## Example
 
 Turbo `webview` can be rendered using native view `VisitableView`.
@@ -161,7 +163,7 @@ webkit.messageHandlers.nativeApp.postMessage(message);
 
 ### `onWebAlert`
 
-Function called when website inside WebView is calling `alert` function. By default React Native's `Alert` is displayed. 
+Function called when website inside WebView is calling `alert` function. By default React Native's `Alert` is displayed.
 
 Note that after handling alert display, `callback` function must be called.
 
--- a/packages/turbo/README.md
+++ b/packages/turbo/README.md
@@ -197,6 +197,10 @@ injectJavaScript(jsCode);
 
 Reloads the webview.
 
+### `clearSnapshotCache`
+
+Clears the snapshot cache. It might be useful in a form submission scenario where different session handles are used for modals: when a form submission is completed in the modal session, we need to manually clear the snapshot cache in the default session to avoid using potentially outdated cached snapshots.
+
 ## Session Component
 
 Session component has been deprecated. To use multiple [sessions](https://github.com/hotwired/turbo-ios/blob/main/Docs/Overview.md#session), you can use `sessionHandle` prop on `VisitableView` component.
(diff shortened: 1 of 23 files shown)


packages/turbo/android/src/main/java/com/reactnativeturbowebview/RNSession.kt
```diff
@@ -120,6 +120,11 @@ class RNSession(
     }
   }
 
+  fun clearSnapshotCache() {
+    // turbo-android doesn't expose a way to clear the snapshot cache, so we have to do it manually
+    webView.evaluateJavascript("window.Turbo.session.clearCache();", null)
```
pfeiffer, line 125: This is a clever plug!

packages/turbo/android/src/main/java/com/reactnativeturbowebview/RNSession.kt
```diff
@@ -154,6 +159,14 @@ class RNSession(
     topmostView?.visitRendered()
   }
 
+  override fun formSubmissionStarted(location: String) {
+    topmostView?.didStartFormSubmission(location)
+  }
+
+  override fun formSubmissionFinished(location: String) {
+    topmostView?.didFinishFormSubmission(location)
```
pfeiffer, line 167: I'm wondering what happens if the location of the `topmostView` has changed in the meantime, ie. due to `url` being updated or a different view has become the topmost one? Is this something we should care about and/or filter?

pklatka: When you change screen during form upload (submissionStarted callback is called and submissionFinished is not) then this new screen (== new topmostView) calls `submissionFinished` callback (it is not called on a screen where form has been submitted). So it appears that sessionDelegate is changed before `formSubmissionFinished` is called. So it might make sense to store what topmostView called `submissionStarted` just to avoid calling `formSubmissionFinished` on a screen which has not submitted a form.

pfeiffer: The corresponding code in `Turbo Navigator iOS` looks like [this](https://github.com/hotwired/turbo-ios/blob/turbo-navigator/Source/Turbo%20Navigator/TurboNavigator.swift#L108-L121):

```swift
    public func sessionDidStartFormSubmission(_ session: Session) {
        if let url = session.topmostVisitable?.visitableURL {
            delegate.formSubmissionDidStart(to: url)
        }
    }

    public func sessionDidFinishFormSubmission(_ session: Session) {
        if session == modalSession {
            self.session.clearSnapshotCache()
        }
        if let url = session.topmostVisitable?.visitableURL {
            delegate.formSubmissionDidFinish(at: url)
        }
    }
```

These seem fairly app-specific (ie. the `session.clearSnapshotCache()` stuff which belongs in the app-code.

I think triggering the `onFormSubmissionFinished` in the `VisitableView` without any filtering is what we want as long as we pass the URL, the corresponding handler can always ignore the call, if the URLs don't match.

We can consider adding some documentation regarding this, ie. that the form submission handlers are triggered for the _session_ and a `url` argument is available for granular control.

packages/turbo/README.md
```diff
@@ -27,6 +27,8 @@ For Android you need to adjust your SDK version in your `build.gradle`.
 >
```
pfeiffer, line 27: Thinking we should add documentation on the `onFormSubmissionStarted` and `onFormSubmissionFinished` as well. We should explain that it happens when the form submission happens on the _session_ and that the `url` argument is available.

packages/turbo/android/src/main/java/com/reactnativeturbowebview/RNVisitableViewManager.kt
```diff
@@ -13,7 +13,9 @@ enum class RNVisitableViewEvent(val jsCallbackName: String) {
   MESSAGE("onMessage"),
   OPEN_EXTERNAL_URL("onOpenExternalUrl"),
   ON_ALERT("onWebAlert"),
-  ON_CONFIRM("onWebConfirm")
+  ON_CONFIRM("onWebConfirm"),
+  FORM_SUBMISSION_START("onFormSubmissionStarted"),
+  FORM_SUBMISSION_FINISH("onFormSubmissionFinished")
```
pfeiffer, line 18: Nitpicking, but there's some inconsistencies in how we name these (`ON_..` vs no prefix and past tense etc).
The inconsistency is there already though. WDYT?

pklatka: Yeah, let's fix that. Because we use this enum in a `sendEvent` let's just remove "on" prefix and keep the enum names synced with corresponding js callback name.

PR #92, opened 2024-01-09
Support defining custom `ActivityIndicator` component

pklatka:
## Summary

This PR adds `renderLoading` prop which allows to define custom `ActivityIndicator` component.

### DIFF
--- a/packages/turbo/README.md
+++ b/packages/turbo/README.md
@@ -124,6 +124,19 @@ const stradaComponents = [FormComponent];
 />
 ```
 
+### `renderLoading`
+
+Function that returns a loading indicator.
+
+Example:
+
+```jsx
+<VisitableView
+  ...
+  renderLoading={() => <Loading />}
+/>
+```
+
 ### `onVisitProposal`
 
 Callback called when the webview detects turbo visit action.
--- a/packages/turbo/README.md
+++ b/packages/turbo/README.md
@@ -126,7 +126,7 @@ const stradaComponents = [FormComponent];
 
 ### `renderLoading`
 
-Function that returns a loading indicator.
+Function that returns a loading indicator, which will be shown when the webview is loading.
 
 Example:
 
--- a/packages/turbo/android/src/main/java/com/reactnativeturbowebview/RNVisitableView.kt
+++ b/packages/turbo/android/src/main/java/com/reactnativeturbowebview/RNVisitableView.kt
@@ -6,6 +6,7 @@ import android.view.ViewGroup
 import android.webkit.CookieManager
 import android.widget.LinearLayout
 import androidx.appcompat.widget.AppCompatImageView
+import androidx.core.view.isVisible
 import androidx.lifecycle.findViewTreeLifecycleOwner
 import com.facebook.react.bridge.Arguments
 import com.facebook.react.bridge.ReactApplicationContext
@@ -222,13 +223,19 @@ class RNVisitableView(context: Context) : LinearLayout(context), SessionSubscrib
   }
 
   private fun showProgressView() {
-    turboView.addProgressView(inflate(context, R.layout.turbo_progress, null))
+    // Don't show the progress view if a screenshot is available
+    if (screenshotView.isVisible) return
+    sendEvent(RNVisitableViewEvent.SHOW_VISITABLE_ACTIVITY_INDICATOR, Arguments.createMap())
+  }
+
+  private fun hideProgressView() {
+    sendEvent(RNVisitableViewEvent.HIDE_VISITABLE_ACTIVITY_INDICATOR, Arguments.createMap())
   }
 
   private fun removeTransitionalViews() {
     turboView.webViewRefresh?.isRefreshing = false
     turboView.errorRefresh?.isRefreshing = false
-    turboView.removeProgressView()
+    hideProgressView()
     turboView.removeScreenshot()
     turboView.removeErrorView()
   }
@@ -324,7 +331,7 @@ class RNVisitableView(context: Context) : LinearLayout(context), SessionSubscrib
     }
   }
 
-  override fun handleAlert(message: String,callback: () -> Unit) {
+  override fun handleAlert(message: String, callback: () -> Unit) {
     sendEvent(RNVisitableViewEvent.WEB_ALERT, Arguments.createMap().apply {
       putString("message", message)
     })
--- a/packages/turbo/android/src/main/java/com/reactnativeturbowebview/RNVisitableView.kt
+++ b/packages/turbo/android/src/main/java/com/reactnativeturbowebview/RNVisitableView.kt
@@ -225,11 +225,11 @@ class RNVisitableView(context: Context) : LinearLayout(context), SessionSubscrib
   private fun showProgressView() {
     // Don't show the progress view if a screenshot is available
     if (screenshotView.isVisible) return
-    sendEvent(RNVisitableViewEvent.SHOW_VISITABLE_ACTIVITY_INDICATOR, Arguments.createMap())
+    sendEvent(RNVisitableViewEvent.SHOW_LOADING, Arguments.createMap())
   }
 
   private fun hideProgressView() {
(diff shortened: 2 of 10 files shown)


packages/turbo/src/hooks/useActivityIndicator.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+export type RenderLoading = () => React.ReactNode;
+
+const defaultRenderLoading = () => (
+  <View style={styles.wrapper}>
+    <ActivityIndicator
+      size={Platform.select({
+        ios: 'small',
+        android: 'large',
+      })}
```
pklatka, line 12: This size is the most similar to turbo-android/iOS native activity indicator.

packages/turbo/src/hooks/useActivityIndicator.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+const styles = StyleSheet.create({
+  wrapper: {
+    position: 'absolute',
+    flex: 1,
+    justifyContent: 'center',
+    alignItems: 'center',
+    height: '100%',
+    width: '100%',
+    backgroundColor: 'white',
```
pfeiffer, line 50: Maybe leave as transparent?

packages/turbo/src/hooks/useActivityIndicator.tsx
pfeiffer, line 12:
```suggestion
const SIZE = Platform.select({
  ios: 'small',
  android: 'large',
})

const defaultRenderLoading = () => (
  <View style={styles.wrapper}>
    <ActivityIndicator
      size={SIZE}
```

packages/turbo/README.md
```diff
@@ -120,6 +120,19 @@ const stradaComponents = [FormComponent];
 />
 ```
 
+### `renderLoading`
+
+Function that returns a loading indicator.
```
pfeiffer, line 125:
```suggestion
Function that returns a loading indicator, which will be shown when the webview is loading.
```

packages/turbo/android/src/main/java/com/reactnativeturbowebview/RNVisitableViewManager.kt
```diff
@@ -15,7 +15,9 @@ enum class RNVisitableViewEvent(val jsCallbackName: String) {
   WEB_ALERT("onWebAlert"),
   WEB_CONFIRM("onWebConfirm"),
   FORM_SUBMISSION_STARTED("onFormSubmissionStarted"),
-  FORM_SUBMISSION_FINISHED("onFormSubmissionFinished")
+  FORM_SUBMISSION_FINISHED("onFormSubmissionFinished"),
+  SHOW_VISITABLE_ACTIVITY_INDICATOR("onShowVisitableActivityIndicator"),
+  HIDE_VISITABLE_ACTIVITY_INDICATOR("onHideVisitableActivityIndicator")
```
pfeiffer, line 20: I think the method names could be improved here. Do we need `Visitable` as part of the name?
There's also a disconnect between `renderLoading` prop name and `..ActivityIndicator..`.

What about `onShowLoading` and on `onHideLoading`?

packages/turbo/src/VisitableView.tsx
```diff
@@ -143,8 +149,15 @@ const VisitableView = React.forwardRef<RefObject, React.PropsWithRef<Props>>(
       onConfirm
     );
 
+    const {
+      activityIndicator,
+      handleShowActivityIndicator,
+      handleHideActivityIndicator,
+    } = useActivityIndicator(renderLoading);
```
pfeiffer, line 156: I think the naming here could also be improved to not refer to `ActivityIndicator`

packages/turbo/src/hooks/useActivityIndicator.tsx
```diff
@@ -0,0 +1,53 @@
...
+    <ActivityIndicator
+      size={Platform.select({
+        ios: 'small',
+        android: 'large',
+      })}
+    />
+  </View>
+);
+
+export function useActivityIndicator(renderLoading?: RenderLoading) {
```
pfeiffer, line 17:
```suggestion
export function useRenderLoading(renderLoading?: RenderLoading) {
```

packages/turbo/src/hooks/useActivityIndicator.tsx
pfeiffer: Okay. I see on Android they add a full view covering the parent with `android:colorBackground` as the background color. For the demo app that's white. I'm okay keeping this for now as it will match both iOS and Android behavior in this case.

packages/turbo/android/src/main/java/com/reactnativeturbowebview/RNVisitableView.kt
```diff
@@ -217,13 +218,19 @@ class RNVisitableView(context: Context) : LinearLayout(context), SessionSubscrib
   }
 
   private fun showProgressView() {
-    turboView.addProgressView(inflate(context, R.layout.turbo_progress, null))
+    // Don't show the progress view if a screenshot is available
+    if (screenshotView.isVisible) return
```
pfeiffer, line 222: Hm, this differs from the [`turbo-android` implementation](https://github.com/hotwired/turbo-android/blob/b5ba0325e3a0ab5e87242a53fea5876d1322f4f5/turbo/src/main/kotlin/dev/hotwire/turbo/delegates/TurboWebFragmentDelegate.kt#L397-L399). Is it intentional?

pfeiffer: If we want this behavior, I feel like the guard clause is better fitted to have in the methods calling this. I find it weird that `showProgressView(..)` can sometimes be a no-op.

packages/turbo/ios/RNVisitableView.swift
```diff
@@ -187,5 +192,14 @@ extension RNVisitableView: RNVisitableViewControllerDelegate {
     ]
     onLoad?(event)
   }
+    
+  func showVisitableActivityIndicator() {
+    guard !isRefreshing else { return }
```
pfeiffer, line 197: Do we need this guard clause?

pklatka: Otherwise we end up displaying loadingComponent when we refresh the webpage. [turbo-ios](https://github.com/hotwired/turbo-ios/blob/c9d060fcdec136c3808cd0764e956d375012c6fb/Source/Visitable/VisitableView.swift#L112-L117) does something similar.

pfeiffer: Allright, makes sense. I think the naming of the method just confuses me a bit then.

packages/turbo/android/src/main/java/com/reactnativeturbowebview/RNVisitableView.kt
pklatka: Yes, the change is intentional. We no longer want to add native loader view to the `turboView`, instead we want to intercept all events responsible for showing a ActivityIindicator. 

This piece of code is also a copy-paste from [turbo-android](https://github.com/hotwired/turbo-android/blob/b5ba0325e3a0ab5e87242a53fea5876d1322f4f5/turbo/src/main/kotlin/dev/hotwire/turbo/views/TurboView.kt#L80-L88).

pfeiffer: Got it! Sounds reasonable and the way to go :-)

pfeiffer: Should we have the guard here for `isRefreshing` as well?

pklatka: A loading animation is triggered in a "refresh spinner":
image

pfeiffer:
LGTM!

PR #94, opened 2024-01-09
Support onContentProcessDidTerminate callback

pklatka:
## Summary

This PR implements `sessionWebViewProcessDidTerminate` callback on iOS and `onRenderProcessGone` on Android and exposes `onContentProcessDidTerminate`, which, by default, reloads webView (similarly to [turbo-ios approach](https://github.com/hotwired/turbo-ios/blob/f4decce15e85218397638b241665604f739e2ba1/Source/Turbo%20Navigator/TurboNavigator.swift#L128-L130))

### DIFF
--- a/packages/turbo/src/VisitableView.tsx
+++ b/packages/turbo/src/VisitableView.tsx
@@ -4,7 +4,7 @@ import React, {
   useRef,
   useMemo,
 } from 'react';
-import { NativeSyntheticEvent, StyleSheet, Platform } from 'react-native';
+import { NativeSyntheticEvent, StyleSheet } from 'react-native';
 import RNVisitableView, {
   dispatchCommand,
   openExternalURL,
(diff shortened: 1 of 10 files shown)


packages/turbo/src/VisitableView.tsx
```diff
@@ -64,6 +66,10 @@ const VisitableView = React.forwardRef<RefObject, React.PropsWithRef<Props>>(
       onConfirm,
       onFormSubmissionStarted,
       onFormSubmissionFinished,
+      onContentProcessDidTerminate = Platform.select({
+        ios: () => dispatchCommand(visitableViewRef, 'reload'),
+        android: () => onVisitProposal({ url, action: 'replace' }),
+      })!,
```
pklatka, line 72: Actually, default behaviours are different on both platforms:
iOS: https://github.com/hotwired/turbo-ios/blob/f4decce15e85218397638b241665604f739e2ba1/Source/Turbo%20Navigator/TurboNavigator.swift#L128-L130
Android: https://github.com/hotwired/turbo-android/blob/b5ba0325e3a0ab5e87242a53fea5876d1322f4f5/turbo/src/main/kotlin/dev/hotwire/turbo/delegates/TurboWebFragmentDelegate.kt#L212-L214

pfeiffer: I'm wondering if we should have 1 method or two separate methods for the platforms. Hm.

Looking at `react-native-webview` they have one method for Android (`onRenderProcessGone`) and one for iOS (`onContentProcessDidTerminate`).

I'd probably vouch for having a single method documenting under which circumstances it's triggered, and having the same behavior (the iOS one of reloading the session).

pfeiffer:
LGTM! Great work 🎉

PR #96, opened 2024-01-10
Enable to define custom Error component

pklatka:
## Summary

This PR:
1. Introduces `renderError` prop which allows to define custom component which will be shown when error occurs. 
2. Introduces default component which is displayed when error occurs
3. Adds `onContentProcessDidTerminate` to documentation
4. Renames `onVisitError` to `onError`

Default error screen:

| Android | iOS |
| -------- | --- |
| Android error | iOS Error |


packages/turbo/src/VisitableView.tsx
```diff
@@ -151,8 +162,16 @@ const VisitableView = React.forwardRef<RefObject, React.PropsWithRef<Props>>(
...
+    const {
+      loadingComponent,
+      handleShowLoading: handleShowLoadingCallback,
+      handleHideLoading,
+    } = useRenderLoading(renderLoading);
+
+    const handleShowLoading = useCallback(() => {
+      hideErrorComponent();
+      handleShowLoadingCallback();
+    }, [handleShowLoadingCallback, hideErrorComponent]);
```
pklatka, line 174: This part of code fixes an issue shown on a video:

https://github.com/software-mansion-labs/react-native-turbo-demo/assets/44744462/5064ce73-fad2-479b-934e-800561c951f0

packages/turbo/src/types.ts
```diff
@@ -27,10 +27,11 @@ export type MessageEvent = object;
 export interface AlertHandler {
   message: string;
 }
-export interface VisitProposalError {
-  statusCode: number;
+
+export interface ErrorEvent {
   url: string;
-  error?: string;
+  statusCode?: number;
```
pklatka, line 33: Status code is not always provided (see iOS native code)

packages/turbo/src/VisitableView.tsx
```diff
@@ -169,6 +188,7 @@ const VisitableView = React.forwardRef<RefObject, React.PropsWithRef<Props>>(
 
     return (
       <>
+        {errorComponent}
         {loadingComponent}
```
pfeiffer, line 192: Thinking we'd probably want to only render one view here (loading should take precedence).

There's a similar pattern in [`react-native-webview`](https://github.com/react-native-webview/react-native-webview/blob/6960a1903c3c32e1e77be11e6370f1c384ed7ee4/src/WebView.ios.tsx#L140-L149C7)

packages/turbo/src/types.ts
```diff
@@ -27,10 +27,11 @@ export type MessageEvent = object;
 export interface AlertHandler {
   message: string;
 }
-export interface VisitProposalError {
-  statusCode: number;
+
+export interface ErrorEvent {
```
pfeiffer, line 31: I think we'd benefit greatly from surfacing the type of the error here as well to allow more granular control. On iOS the `TurboError` is an enum, so we should be able to be more specific about the type here.

pfeiffer: In theory we could always return the statusCode, which would then be `0` for `networkFailure` etc.

packages/turbo/ios/RNVisitableView.swift
```diff
@@ -126,16 +126,31 @@ class RNVisitableView: UIView, RNSessionSubscriber {
       onVisitProposal!(event)
     }
   }
+    
+  func getStatusCodeFromError(error: TurboError?) -> Int {
```
pklatka, line 130: I'm unable to extend `TurboError` enum, so I couldn't match Android style (I also wanted to avoid maintaining two identical enums on iOS, it would be better to inherit from it but it's impossible)

packages/turbo/README.md
```diff
@@ -38,15 +38,14 @@ import {
...
-  VisitProposalError,
   Session,
 } from 'react-native-turbo';
 import { useNavigation } from '@react-navigation/native';
 
 const TurboScreen = () => {
   const navigation = useNavigation();
 
-  const onVisitProposal = ({ nativeEvent: { action: actionType, url } }) => {
+  const onVisitProposal = ({ action: actionType, url }) => {
```
tjzel, line 48: That's a very good change here!

packages/turbo/android/src/main/java/com/reactnativeturbowebview/RNTurboError.kt
```diff
@@ -0,0 +1,50 @@
+package com.reactnativeturbowebview
+
+import android.webkit.WebViewClient
+
+enum class RNTurboError(val code: Int) {
+  HTTP(1),
+  NETWORK_FAILURE(0),
```
tjzel, line 7: Just out of curiosity, I wonder if you really want to have `0` as an error code, since in general (and in POSIX) `0` means success.

tjzel (changes requested):
TypeScript looks solid in general, I just have a few questions and a few things could use some more polishing.

packages/turbo/android/src/main/java/com/reactnativeturbowebview/RNTurboError.kt
pklatka: Unfortunately `turbo-ios` defines code 0 as a network failure so to match both native implementations I had to use this code (see [this](https://github.com/hotwired/turbo-ios/blob/c9d060fcdec136c3808cd0764e956d375012c6fb/Source/TurboError.swift#L4))

packages/turbo/src/VisitableView.tsx
```diff
@@ -98,16 +115,17 @@ const VisitableView = React.forwardRef<RefObject, React.PropsWithRef<Props>>(
...
+        reload: reloadVisitableView,
       }),
-      []
+      [reloadVisitableView]
     );
 
-    const handleVisitError = useCallback(
-      ({ nativeEvent }: NativeSyntheticEvent<VisitProposalError>) => {
-        viewErrorHandler?.(nativeEvent);
+    const handleOnError = useCallback(
```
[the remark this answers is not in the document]
pklatka: Makes sense!

packages/turbo/src/hooks/useWebViewState.tsx
```diff
@@ -0,0 +1,113 @@
...
+  Button,
+  ActivityIndicator,
+  View,
+  StyleSheet,
+  Platform,
+} from 'react-native';
+import { ErrorEvent } from '../types';
+
+export type RenderLoading = () => React.ReactNode;
+export type RenderError = () => React.ReactNode;
```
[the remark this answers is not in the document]
pklatka: Good catch!

packages/turbo/android/src/main/java/com/reactnativeturbowebview/RNTurboError.kt
tjzel: 😢

example/src/MainScreen.tsx
```diff
@@ -0,0 +1,24 @@
...
+        props.navigation.navigate(Routes.SignIn, { path: 'signin' });
+      } else {
+        setRenderError(undefined);
+      }
+    },
+    [props.navigation]
+  );
+
+  return <WebView {...props} onError={onError} renderError={renderError} />;
+};
```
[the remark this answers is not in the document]
tjzel: I see. Just change the type of `useState<any>` to `useState<RenderError | undefined>` then and all is good!

PR #97, opened 2024-01-11
Get correct key when payload has nested params

### LINKED ISSUE #88: Issue with nested navigators and `useWebviewNavigate`
There is an issue with how `useWebviewNavigate()` works with nested navigators. In `getAction` a `key` is added to uniquely identify a screen to be pushed. This assumes that the `action.payload` is always a flat object, eg:

```js
{
  "type": "NAVIGATE",
  "payload": {
     "name": "SomeScreen",
     "params": { .. }
  }
}
```

However, for nested navigators, the payload can look like this:

```js
{
  "type": "NAVIGATE",
  "payload": {
    "name": "FocusedFlow",
    "params": {
      "initial": true,
      "screen": "Subscription",
      "params": {
        "step": "checkout",
        "baseURL": "https://.../",
        "fullPath": "/subscribe/checkout"
      },
      "path": "/subscribe/checkout"
    }
  }
}
```

... in which case the `fullPath` will be tried extracted from the top-level object and not the Subscription screen in this case. We want to add the `key` and extract the `fullPath` in the deepest screen.

In practice, this causes screens to even not be pushed in case of a fallback match (as they have an undefined `key`) or we might push certain screens multiple times (in this case the `FocusedFlow` is a modal and navigating in that will open a new modal on top of the previous one instead of pushing the screen in the nested stack navigator)

mironiasty:
Fixes #88

Use recursion to read `fullPath` from `payload` object. This solves issue when `fullPath` is not read on nested navigation.

### DIFF
--- a/packages/navigation/src/hooks/useWebviewNavigate.ts
+++ b/packages/navigation/src/hooks/useWebviewNavigate.ts
@@ -46,6 +46,17 @@ function isNavigateAction(
   return !!action && action.type === 'NAVIGATE';
 }
 
+function getKeyFromParams(params: unknown) {
+  if (params && typeof params === 'object') {
+    if ('fullPath' in params && params.fullPath) {
+      return params.fullPath as string;
+    } else if ('params' in params) {
+      return getKeyFromParams(params.params);
+    }
+  }
+  return undefined;
+}
+
 function getAction(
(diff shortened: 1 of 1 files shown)


pfeiffer:
LGTM!

PR #99, opened 2024-01-11
Always call completion handler for alert/confirm dialog

### LINKED ISSUE #95: iOS: Crash when an alert/confirm is showing in a non-topmost WebView
We added support for showing `alert(..)` and `confirm(..)` in #76. During my testing, I've seen a a consistent reproducible crash when showing alerts in more than one webview:

```
NSInternalInconsistencyException: Completion handler passed to -[RNTurbo.TurboUIDelegate webView:runJavaScriptAlertPanelWithMessage:initiatedByFrame:completionHandler:] was not called
  0   CoreFoundation                      0x000000010dc1e330 __exceptionPreprocess + 172
  1   libobjc.A.dylib                     0x00000001075dd274 objc_exception_throw + 56
  2   CoreFoundation                      0x000000010dc1e240 -[NSException initWithCoder:] + 0
  3   WebKit                              0x0000000112f3e6e8 _ZN6WebKit28CompletionHandlerCallCheckerD2Ev + 140
  4   WebKit                              0x0000000112fb4b6c _ZNK3WTF20ThreadSafeRefCountedIN6WebKit28CompletionHandlerCallCheckerELNS_17DestructionThreadE0EE5derefEv + 40
  5   WebKit                              0x0000000113134778 _ZZN3WTF8BlockPtrIFvvEE12fromCallableIZN6WebKit10UIDelegate8UIClient18runJavaScriptAlertERNS4_12WebPageProxyERKNS_6StringEPNS4_13WebFrameProxyEONS4_13FrameInfoDataEONS_8FunctionIS1_EEE3$_4EES2_T_ENUlPKvE_8__invokeESM_ + 32
```

Reproducable case:

1. Have a Tab navigator with eg. 3 tabs
2. Add an `alert(..)` on the web pages shown by the tabs (or use the same URL for multiple tab screens)
3. Navigate to a tab, but before the `alert(..)` is shown, navigate back to the previous one
4. App crashes as the now-hidden tab is loaded

The same thing happens for `confirm(..)`. The tabs use different session handles.

mironiasty:
Fixes #95 

If `visitableView` was not present and we handled `alert`/`confirm` dialog from website -> callbacks methods were never called.
This RP fixes in on both platforms. On iOS this caused app crash (see #95). On Android it caused 'just webview' crash -> website was no longer visible after `alert` method was called.

### DIFF
--- a/packages/turbo/android/src/main/java/com/reactnativeturbowebview/RNWebChromeClient.kt
+++ b/packages/turbo/android/src/main/java/com/reactnativeturbowebview/RNWebChromeClient.kt
@@ -58,16 +58,25 @@ class RNWebChromeClient(
   override fun onJsAlert(
     view: WebView?, url: String?, message: String?, result: JsResult?
   ): Boolean {
-    visitableViews.lastOrNull()?.handleAlert(message ?: "") { result?.confirm() }
+    visitableViews.lastOrNull()?.let {
+      it.handleAlert(message ?: "") { result?.confirm() }
+    } ?: run {
+      result?.confirm()
+    }
+
     return true
   }
 
   override fun onJsConfirm(
     view: WebView?, url: String?, message: String?, result: JsResult?
   ): Boolean {
-    visitableViews.lastOrNull()?.handleConfirm(
-      message ?: ""
-    ) { confirmed -> if (confirmed) result?.confirm() else result?.cancel() }
+    visitableViews.lastOrNull()?.let {
+      it.handleConfirm(
+        message ?: ""
+      ) { confirmed -> if (confirmed) result?.confirm() else result?.cancel() }
+    } ?: run {
+      result?.cancel()
+    }
     return true
   }
 }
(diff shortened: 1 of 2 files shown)


packages/turbo/android/src/main/java/com/reactnativeturbowebview/RNWebChromeClient.kt
```diff
@@ -58,16 +58,25 @@ class RNWebChromeClient(
   override fun onJsAlert(
     view: WebView?, url: String?, message: String?, result: JsResult?
   ): Boolean {
-    visitableViews.lastOrNull()?.handleAlert(message ?: "") { result?.confirm() }
+    visitableViews.lastOrNull()?.let {
+      it.handleAlert(message ?: "") { result?.confirm() }
+    } ?: run {
+      result?.confirm()
+    }
+
```
mironiasty, line 66:
```suggestion

```

pklatka (approved):
Makes sense, LGTM!

PR #101, opened 2024-01-19
Remove @ts-expect-error in RNVisitableView

pklatka:
## Summary

Thanks to @tjzel, this PR removes `@ts-expect-error` from RNVisitableView.


tjzel (approved):
👍

PR #103, opened 2024-01-23
getMinimalActions tweaks

pfeiffer:
Fixes the issues found during testing, namely:

- Avoid infinite loop in `getMinimalAction`
- Adds ability to navigate back in a stack to an already existing named screen/URL
- Replace actions in nested navigators are now working
- Pushing screens w/same name but different URLs (eg. `Fallback`) now work properly again after removal of `key`


packages/navigation/src/hooks/useWebviewNavigate.ts
```diff
@@ -105,6 +105,9 @@ function getMinimalAction(
   while (
     currentAction.payload &&
     'name' in currentAction.payload &&
+    currentAction.payload.name &&
+    'params' in currentAction.payload &&
+    currentAction.payload.params.screen &&
```
pfeiffer, line 110: This fixes issue where the navigating to the same named screen in a stack would dispatch an empty navigation payload (`{}`) as all screens were equal and thus removed.

packages/navigation/src/hooks/useWebviewNavigate.ts
```diff
@@ -105,6 +105,9 @@ function getMinimalAction(
   while (
     currentAction.payload &&
     'name' in currentAction.payload &&
+    currentAction.payload.name &&
```
pfeiffer, line 108: Fixes infinite loop as `name` could be `undefined` and `currentState?.routes[currentState.index ?? -1]?.name` could also be undefined.

packages/navigation/src/hooks/useWebviewNavigate.ts
```diff
@@ -160,10 +163,7 @@ export function useWebviewNavigate<
         : getStateFromPath(path, options?.config);
 
       if (state) {
-        // for REPLACE action we need to pass only top of navigation state.
-        const actionState =
-          actionType === 'replace' ? unpackState(state) : state;
```
pfeiffer, line 165: No need to unpack state as the `getMinimalAction` essentially does the same, but supporting nested structures correctly.

example/src/App.tsx
```diff
@@ -55,10 +56,18 @@ function Account() {
 function FocusedFlow() {
   return (
     <Stack.Navigator screenOptions={{ title: '' }}>
-      <Stack.Screen name={Routes.PhoneActivation} component={WebView} />
-      <Stack.Screen name={Routes.New} component={WebView} />
-      <Stack.Screen name={Routes.SuccessScreen} component={WebView} />
-      <Stack.Screen name={Routes.Fallback} component={WebView} />
+      <Stack.Screen
+        name={Routes.PhoneActivation}
+        getId={getId}
```
pfeiffer, line 61: Needed as we've removed the `key` from navigation payload. `key` is also removed in React Navigation 8.0.

Would love to explore a different pattern for a custom navigator (eg. `const Stack = createWebStackNavigator(..)`) so that this wouldn't be needed to be set explicitly for all defined screens.

packages/navigation/src/hooks/useWebviewNavigate.ts
pfeiffer: `unpackState` import and utility method should be removed in this case

packages/navigation/src/hooks/useWebviewNavigate.ts
```diff
@@ -105,6 +105,9 @@ function getMinimalAction(
   while (
     currentAction.payload &&
```
pfeiffer, line 106: We still need to compare the params of the screen to ensure that we do not remove screens which have same name, but different params as discussed on Slack

pfeiffer:
Superseeded by #106

PR #104, opened 2024-01-23
VisitableView: Remount Strada component when URL changes

pfeiffer:
This PR adds support for remounting Strada components as the URL is changed. This is important for screens that use `replace` where the params are updated (and via that the `url` prop), but the `VisitableView` stays the same.


pklatka (approved):
👍

PR #105, opened 2024-01-24
Android: use assets from turbo-android

pklatka:
## Summary

This PR imports the `R` class from `turbo-android` to correctly identify native views provided by that library. While testing the library in Expo SDK 50 app, I discovered that Android fails to build because it searches for turbo constants in `react-native-turbo` instead of `turbo-android`.


pklatka:
👀

PR #107, opened 2024-01-31
iOS: Refactor and fix visitable delegates

pfeiffer:
This PR refactors the iOS delegates and addresses a few issues:

1. Previously, we had an array of all registered visitableViews in the session, primarily to support modals in the same session. This is not recommended/supported by Turbo iOS, so we can drop support for this, which simplifies the code quite a lot.
2. We did not notify the `TurboSession` via `viewWillAppear` previously; Turbo contains a few optimizations to complete navigations while a forward/back animation is in-flight, so calling this now fixes that
3. The `RNVisitableViewController` inherits from `VisitableViewController` which already has delegates for `visitableViewWillAppear` and calling that on the TurboSession. We remove the explicit call here to avoid double-calling it.

### DIFF
--- a/packages/turbo/ios/RNSession.swift
+++ b/packages/turbo/ios/RNSession.swift
@@ -68,6 +68,10 @@ class RNSession: NSObject {
   }
   
   func visitableViewDidAppear(view: RNVisitableView) {
+    // if (visitableView !== nil && visitableView.isModal !== view.isModal) {
+    //   print("You're not able to share session between modal and non-modals.")
+    // }
+    
     self.visitableView = view
   }
   
--- a/packages/turbo/ios/RNSession.swift
+++ b/packages/turbo/ios/RNSession.swift
@@ -32,7 +32,7 @@ class TurboUIDelegate: NSObject, WKUIDelegate {
 
 class RNSession: NSObject {
   
-  private var visitableViews: [RNSessionSubscriber] = []
+  private var visitableView: RNVisitableView?
   private var sessionHandle: NSString
   private var webViewConfiguration: WKWebViewConfiguration
   private var wkUiDelegate: WKUIDelegate?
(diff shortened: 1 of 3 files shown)


packages/turbo/ios/RNSession.swift
```diff
@@ -61,102 +61,86 @@ class RNSession: NSObject {
...
-      guard let newView = visitableViews.last else {
-        return
-      }
-      visitableViewWillAppear(view: newView)
-    }
-  }
+  func visitableViewDidAppear(view: RNVisitableView) {
+    // if (visitableView !== nil && visitableView.isModal !== view.isModal) {
+    //   print("You're not able to share session between modal and non-modals.")
+    // }
```
pfeiffer, line 73: Not sure if we should address this in the PR?

packages/turbo/ios/RNVisitableView.swift
```diff
@@ -56,10 +54,13 @@ class RNVisitableView: UIView, RNSessionSubscriber {
     return controller
   }()
     
-  private var viewIsRegistered = false
   private var isRefreshing: Bool {
     controller.visitableView.isRefreshing
   }
+  
+  // var isModal: Bool {
+  //   return controller.reactViewController()?.isModal()
```
pklatka, line 62: Xcode tells me that this is impossible to call...

packages/turbo/ios/RNSession.swift
pklatka: We can cover that in this PR since it is breaking previous behaviour but I would prefer to state that in a docs rather than throwing/printing an error on the native side.

pklatka (commented):
Follow-up PR which fixes example app: #109

pklatka (approved):
Merging, we'll figure out the `isModal` property in a follow-up PR

PR #112, opened 2024-02-08
Modal flow: introduce `getDispatchAction` function

pklatka:
This PR introduces a notable change in the `useWebviewNavigate` hook - from now we will return `linkTo` function as well as `getDispatchAction` function. We want to allow to define custom behaviour of `linkTo` function.


pfeiffer:
Could we update the title/description of this PR to explicitly note that it changes behaviors around modals? I think we'll have great use of that in the merge history :-)

PR #113, opened 2024-02-08
Sending visibility event to the website

pklatka:


packages/turbo/android/src/main/java/com/reactnativeturbowebview/RNVisitableView.kt
```diff
@@ -172,14 +172,17 @@ class RNVisitableView(context: Context) : LinearLayout(context), SessionSubscrib
 
     turboView.attachWebView(webView) { attachedToNewDestination ->
       onReady(attachedToNewDestination)
+      sendEvent(RNVisitableViewEvent.WEB_VIEW_MOUNTED, Arguments.createMap())
     }
   }
 
   override fun detachWebView() {
+    sendEvent(RNVisitableViewEvent.WEB_VIEW_UNMOUNTED, Arguments.createMap())
```
pklatka, line 180: Sent too late on Android. iOS has willDisappear callback and it's working properly for now. We can probably handle that event in a return statement in `useEffect` but maybe it will be more convenient to introduce another prop? I'm afraid about potential race condition though.

example/src/WebView.tsx
```diff
@@ -36,15 +52,26 @@ const WebView: React.FC<Props> = ({ navigation, ...props }) => {
...
   );
 
+  const onWebViewMount = useCallback(() => {
+    setWebViewVisibilityState(true);
+  }, [setWebViewVisibilityState]);
+
+  const onWebViewUnmount = useCallback(() => {
+    setWebViewVisibilityState(false);
+  }, [setWebViewVisibilityState]);
+
```
pklatka, line 62: Previous code in `useFocusEffect` was being executed to early - `sessionHandle` prop (at least on iOS) was not initialised and caused crash. Introducing props which make sure that native side is properly mounted might be a better way to go.

pfeiffer:
Let's discuss this in a call; I think this turned out to be way more complex and than it needed and not sure it's worth it.

PR #115, opened 2024-02-10
Bugfix: Make `useCurrentUrl` work properly in top tab navigators

pfeiffer:
This PR fixes a bug where `useCurrentUrl` returned a wrong result if used in tab navigator, where multiple screens would be mounted at the same time (ie. `lazy: false` or if swiping between two screens).

Previously the route params etc. were extracted in a very manual manner using `useNavigation()` and based on the `state.index`. Since multiple screens in the tab navigator would be mounted at the same time, while the `state.index` is static (until actually navigating), we returned the URL from one screen (the one corresponding to `state.index`) for all screens.

The solution is to use the built-in hook `useRoute()` to return the route for a screen.

### DIFF
--- a/packages/navigation/src/hooks/useCurrentUrl.ts
+++ b/packages/navigation/src/hooks/useCurrentUrl.ts
@@ -1,4 +1,4 @@
-import { LinkingOptions, useNavigation } from '@react-navigation/native';
+import { LinkingOptions, useRoute } from '@react-navigation/native';
 
 export type LinkingConfig = LinkingOptions<{}>['config'];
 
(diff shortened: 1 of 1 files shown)


pfeiffer:
TBH I'm think a similar issue would be the case for `useSessionHandle()` which also uses `useNavigation` and `state.index`: https://github.com/software-mansion-labs/react-native-turbo-demo/blob/main/example/src/useSessionHandle.ts#L17-L22

This would lead to unexpected session handles in situations where a navigator can have multiple screens mounted at the same time.

pklatka (approved):
🎉

PR #116, opened 2024-02-10
iOS: Fix visitProposal being proposed before a visitableView is set

pfeiffer:
This PR fixes a "race condition" bug that happens when a visit is instantly proposed from a JavaScript visit (ie. the loaded page issues a `Turbo.visit(..)` on load.

In this case, the `visitableView` on the session would not be set before the new visit is proposed. This causes the proposal to be ignored, as there's no `visitableView` to notify.

Turbo iOS performs visit completion as part of the `..WillAppear(..)` flow, so this PR updates the place where we set the `visitableView` in `RNSession` to match that.

### DIFF
--- a/packages/turbo/ios/RNSession.swift
+++ b/packages/turbo/ios/RNSession.swift
@@ -67,12 +67,14 @@ class RNSession: NSObject {
     }
   }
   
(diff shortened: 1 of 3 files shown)


pklatka (approved):
👍

PR #117, opened 2024-02-14
Refactor refresh/reload utilities

pklatka:
## Summary

This PR:
1. Introduces `getSessionHandles()` method which returns sessionHandles which are available on the native side
2. Introduces `reloadSession(sessionHandle)` method which reloads WebView for given sessionHandle
3. Introduces `refreshSession(sessionHandle)` method which issues a replace visit for given sessionHandle
4. Refactors `clearSnapshotCacheForAllSessions()` to `clearSessionSnapshotCache(sessionHandle)`
5. Properly issues a replace visit in `didProposeVisit` function on iOS
6. Introduces `refresh` method on `VisitableView` component

### DIFF
--- a/packages/turbo/README.md
+++ b/packages/turbo/README.md
@@ -252,14 +252,14 @@ Session component has been deprecated. To use multiple [sessions](https://github
 
 ## Other utilities
 
-### `getRegisteredSessionHandles`
+### `getSessionHandles`
 
 Returns an array of all registered session handles.
 
-### `reloadSessionByName`
+### `reloadSession`
 
 Reloads the webview for given `sessionHandle`.
 
-### `clearSessionSnapshotCacheByName`
+### `clearSessionSnapshotCache`
 
 Clears the snapshot cache for given `sessionHandle`. This might be useful in a scenario where different session handles are used for modals. When a form submission is completed in the modal session, we need to manually clear the snapshot cache in the default session to avoid the use of potentially outdated cached snapshots.
--- a/packages/turbo/README.md
+++ b/packages/turbo/README.md
@@ -244,7 +244,11 @@ injectJavaScript(jsCode);
 
 ### `reload()`
 
-Reloads the webview.
+Reloads the webview. This method is equivalent to the WKWebView's `reload` method.
+
+### `refresh()`
+
+Refreshes the webview. This method is equivalent to making a turbo replace visit to the current URL.
 
 ## Session Component
 
@@ -252,14 +256,20 @@ Session component has been deprecated. To use multiple [sessions](https://github
 
 ## Other utilities
 
+Functions below might be useful in a scenario when a form submission is completed in the modal session - we need to manually clear the snapshot cache in the default session to avoid the use of potentially outdated cached snapshots.
+
 ### `getSessionHandles()`
 
 Returns an array of all registered session handles.
 
 ### `reloadSession(sessionHandle)`
 
-Reloads the webview for given `sessionHandle`.
+Reloads the webview for given `sessionHandle`. This method is equivalent to the WKWebView's `reload` method.
+
+### `refreshSession(sessionHandle)`
+
+Refreshes the webview for given `sessionHandle`. This method is equivalent to making a turbo replace visit to the current URL.
 
 ### `clearSessionSnapshotCache(sessionHandle)`
 
-Clears the snapshot cache for given `sessionHandle`. This might be useful in a scenario where different session handles are used for modals. When a form submission is completed in the modal session, we need to manually clear the snapshot cache in the default session to avoid the use of potentially outdated cached snapshots.
+Clears the snapshot cache for given `sessionHandle`.
--- a/packages/turbo/README.md
+++ b/packages/turbo/README.md
@@ -252,14 +252,14 @@ Session component has been deprecated. To use multiple [sessions](https://github
 
 ## Other utilities
 
-### `getSessionHandles`
+### `getSessionHandles()`
 
 Returns an array of all registered session handles.
 
-### `reloadSession`
+### `reloadSession(sessionHandle)`
 
 Reloads the webview for given `sessionHandle`.
 
-### `clearSessionSnapshotCache`
+### `clearSessionSnapshotCache(sessionHandle)`
 
 Clears the snapshot cache for given `sessionHandle`. This might be useful in a scenario where different session handles are used for modals. When a form submission is completed in the modal session, we need to manually clear the snapshot cache in the default session to avoid the use of potentially outdated cached snapshots.
--- a/packages/turbo/README.md
+++ b/packages/turbo/README.md
@@ -244,11 +244,11 @@ injectJavaScript(jsCode);
 
 ### `reload()`
 
-Reloads the webview. This method is equivalent to the WKWebView's `reload` method.
+Reloads the webview.
 
 ### `refresh()`
 
-Refreshes the webview. This method is equivalent to making a turbo replace visit to the current URL.
+Refreshes the page. This method is equivalent to making a Turbo `replace` visit to the current URL.
 
 ## Session Component
 
@@ -264,12 +264,12 @@ Returns an array of all registered session handles.
 
 ### `reloadSession(sessionHandle)`
 
-Reloads the webview for given `sessionHandle`. This method is equivalent to the WKWebView's `reload` method.
+Reloads the page for the given `sessionHandle`.
 
 ### `refreshSession(sessionHandle)`
 
-Refreshes the webview for given `sessionHandle`. This method is equivalent to making a turbo replace visit to the current URL.
+Refreshes the page for the given `sessionHandle`. This method is equivalent to making a Turbo `replace` visit to the current URL.
 
 ### `clearSessionSnapshotCache(sessionHandle)`
 
-Clears the snapshot cache for given `sessionHandle`.
+Clears the snapshot cache for the given `sessionHandle`.
--- a/packages/turbo/README.md
+++ b/packages/turbo/README.md
@@ -252,6 +252,14 @@ Session component has been deprecated. To use multiple [sessions](https://github
 
 ## Other utilities
 
-### `clearSnapshotCacheForAllSessions`
+### `getRegisteredSessionHandles`
 
-Clears the snapshot cache for all sessions. This might be useful in a scenario where different session handles are used for modals. When a form submission is completed in the modal session, we need to manually clear the snapshot cache in the default session to avoid the use of potentially outdated cached snapshots.
+Returns an array of all registered session handles.
+
+### `reloadSessionByName`
+
+Reloads the webview for given `sessionHandle`.
+
+### `clearSessionSnapshotCacheByName`
+
+Clears the snapshot cache for given `sessionHandle`. This might be useful in a scenario where different session handles are used for modals. When a form submission is completed in the modal session, we need to manually clear the snapshot cache in the default session to avoid the use of potentially outdated cached snapshots.
--- a/packages/turbo/android/src/main/java/com/reactnativeturbowebview/RNVisitableView.kt
+++ b/packages/turbo/android/src/main/java/com/reactnativeturbowebview/RNVisitableView.kt
@@ -63,7 +63,6 @@ class RNVisitableView(context: Context) : LinearLayout(context), SessionSubscrib
   private var screenshotOrientation = 0
   private var screenshotZoomed = false
   private var screenshot: Bitmap? = null
-  private var hideActivityIndicatorOnNextLoad = false
 
   // Views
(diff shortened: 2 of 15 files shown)


packages/turbo/README.md
```diff
@@ -252,6 +252,14 @@ Session component has been deprecated. To use multiple [sessions](https://github
 
 ## Other utilities
 
-### `clearSnapshotCacheForAllSessions`
+### `getRegisteredSessionHandles`
```
pfeiffer, line 255: Nitpick, but is `Registered` needed here? We don't have any session handles that are not registered.

packages/turbo/README.md
```diff
@@ -252,6 +252,14 @@ Session component has been deprecated. To use multiple [sessions](https://github
...
 
-Clears the snapshot cache for all sessions. This might be useful in a scenario where different session handles are used for modals. When a form submission is completed in the modal session, we need to manually clear the snapshot cache in the default session to avoid the use of potentially outdated cached snapshots.
+Returns an array of all registered session handles.
+
+### `reloadSessionByName`
```
pfeiffer, line 259: We could consider dropping `ByName` here, so signature would be `reloadSession(name)`?

packages/turbo/README.md
```diff
@@ -252,6 +252,14 @@ Session component has been deprecated. To use multiple [sessions](https://github
...
+
+Reloads the webview for given `sessionHandle`.
+
+### `clearSessionSnapshotCacheByName`
```
pfeiffer, line 263: Same as above; `clearSessionSnapshotCache(..)`

I find `byName` confusing here - we call it `sessionHandle` elsewhere. I find it clearer that we just drop the by name.

packages/turbo/README.md
pklatka: Makes sense

packages/turbo/ios/RNVisitableView.swift
```diff
@@ -47,69 +47,73 @@ class RNVisitableView: UIView, RNSessionSubscriber {
...
-  
+
   public func sendConfirmResult(result: NSString) -> Void {
     let confirmResult = result == "true"
     self.onConfirmHandler?(confirmResult)
     self.onConfirmHandler = nil
   }
-  
+
   public func reload(){
```
pfeiffer, line 93:
```suggestion
  public func reload() {
```

packages/turbo/ios/RNSession.swift
```diff
@@ -80,8 +80,12 @@ class RNSession: NSObject {
   func visit(_ visitable: Visitable) {
     turboSession.visit(visitable)
   }
-  
+    
   func reload() {
+    webView.reload()
+  }
+  
+  func refresh() {
```
pfeiffer, line 88: This is not what I think we want. We want the `refresh` to perform a `replace` visit to the current URL, and the `reload` to reload the `turboSession`

pfeiffer: Ie. something along the lines of:

```swift
func reload() {
  turboSession.reload()
}

func refresh() {
  // turboSession.visit with the current URL and action = replace
}

File: packages/turbo/android/src/main/java/com/reactnativeturbowebview/RNVisitableViewManager.kt [hunk 6]
```diff
@@ -70,7 +71,8 @@ class RNVisitableViewManager(
           root.injectJavaScript(it)
         }
       }
-      RNVisitableViewCommand.RELOAD_PAGE -> root.refresh(true)
+      RNVisitableViewCommand.RELOAD_PAGE -> root.reload()
+      RNVisitableViewCommand.REFRESH_PAGE -> root.refresh(true)
```
pfeiffer, line 75: Same as iOS here. We do not want to show progress when _refreshing_ only when _reloading_.

packages/turbo/ios/RNSession.swift
pklatka: `turboSession.reload()` issues a visit with `topmostVisit` as a parameter:
```
    public func reload() {
        guard let visitable = topmostVisitable else { return }

        initialized = false
        visit(visitable)
        topmostVisit = currentVisit
    }
   ```

I'm wondering if that's what we want - we want to do a coldboot not another visit so let's go with:

```
func reload() {
  webView.reload()
}

func refresh() {
  // turboSession.visit with the current URL and action = replace
}
```

Comment (pfeiffer) [reply packages/turbo/ios/RNSession.swift:88]: I'm pretty sure the `turboSession.reload()` is what we want. It reloads the current page setting `initialized = false` before, which issues a `ColdBoot` visit:

image

File: packages/turbo/README.md [hunk 7]
```diff
@@ -244,14 +244,32 @@ injectJavaScript(jsCode);
 
 ### `reload()`
 
-Reloads the webview.
+Reloads the webview. This method is equivalent to the WKWebView's `reload` method.
```
pfeiffer, line 247: Nitpicking, but we `reload` the Turbo Session here, not the WebView.

packages/turbo/README.md
```diff
@@ -244,14 +244,32 @@ injectJavaScript(jsCode);
...
+
+### `refresh()`
+
+Refreshes the webview. This method is equivalent to making a turbo replace visit to the current URL.
```
pfeiffer, line 251:
```suggestion
Refreshes the page. This method is equivalent to making a Turbo `replace` visit to the current URL.
```

packages/turbo/README.md
```diff
@@ -244,14 +244,32 @@ injectJavaScript(jsCode);
...
+
+Returns an array of all registered session handles.
+
+### `reloadSession(sessionHandle)`
+
+Reloads the webview for given `sessionHandle`. This method is equivalent to the WKWebView's `reload` method.
+
+### `refreshSession(sessionHandle)`
+
+Refreshes the webview for given `sessionHandle`. This method is equivalent to making a turbo replace visit to the current URL.
```
pfeiffer, line 271:
```suggestion
Refreshes the page for given `sessionHandle`. This method is equivalent to making a Turbo `replace` visit to the current URL.
```

packages/turbo/README.md
```diff
@@ -244,14 +244,32 @@ injectJavaScript(jsCode);
...
-### `clearSnapshotCacheForAllSessions`
+Functions below might be useful in a scenario when a form submission is completed in the modal session - we need to manually clear the snapshot cache in the default session to avoid the use of potentially outdated cached snapshots.
+
+### `getSessionHandles()`
+
+Returns an array of all registered session handles.
+
+### `reloadSession(sessionHandle)`
+
+Reloads the webview for given `sessionHandle`. This method is equivalent to the WKWebView's `reload` method.
```
pfeiffer, line 267:
```suggestion
Reloads the page for given `sessionHandle`.
```

packages/turbo/README.md
```diff
@@ -244,14 +244,32 @@ injectJavaScript(jsCode);
...
+
+### `refreshSession(sessionHandle)`
+
+Refreshes the webview for given `sessionHandle`. This method is equivalent to making a turbo replace visit to the current URL.
+
+### `clearSessionSnapshotCache(sessionHandle)`
 
-Clears the snapshot cache for all sessions. This might be useful in a scenario where different session handles are used for modals. When a form submission is completed in the modal session, we need to manually clear the snapshot cache in the default session to avoid the use of potentially outdated cached snapshots.
+Clears the snapshot cache for given `sessionHandle`.
```
pfeiffer, line 275:
```suggestion
Clears the snapshot cache for the given `sessionHandle`.
```

packages/turbo/android/src/main/java/com/reactnativeturbowebview/RNVisitableView.kt
```diff
@@ -109,7 +109,11 @@ class RNVisitableView(context: Context) : LinearLayout(context), SessionSubscrib
     }
   }
 
-  internal fun refresh(displayProgress: Boolean) {
+  fun reload() {
+    session.reload()
+  }
+
+  override fun refresh(displayProgress: Boolean) {
```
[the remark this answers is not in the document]
pfeiffer: Something like:

```kt
fun reload(displayProgress: Boolean)
  // .. probably the code we have in `refresh` now or simply implement and call
  // session.reload(displayProgress: ..)
}

override fun refresh() {
  session.refresh()
}
```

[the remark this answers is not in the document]
pfeiffer: I understand that previously `refresh` (on Android) was referring to what we now call `reload`.

We have however implemented most of Turbo Android on our own, so I think it's better that we clear this up and have parity in naming and functionality with iOS

packages/turbo/android/src/main/java/com/reactnativeturbowebview/RNVisitableView.kt
```diff
@@ -122,18 +134,18 @@ class RNVisitableView(context: Context) : LinearLayout(context), SessionSubscrib
     performVisit(restoreWithCachedSnapshot = false, reload = true)
   }
 
-  private fun initializePullToRefresh(turboView: TurboView) {
+  private fun initializePullToReload(turboView: TurboView) {
```
pfeiffer, line 137: Is we rename this, we might as well also rename `isRefreshing` to `isReloading`.

pfeiffer: Or maybe not. Not sure. What do you think? It's somewhat confusing that the Android uses `setOnRefreshListener` here.

Hm.

I mean; `setOnRefreshListener` refers to the Android UI refreshing action, which _reloads_ our session. We then also have a separate `refresh(..)` method that is totally unrelated (and something else) than `reload()` which refreshes the session's page.

pfeiffer: I think we should keep the old names, `initializePullToRefresh` etc. and not rename those.

packages/turbo/android/src/main/java/com/reactnativeturbowebview/RNVisitableView.kt
```diff
@@ -39,7 +40,7 @@ class RNVisitableView(context: Context) : LinearLayout(context), SessionSubscrib
   var pullToRefreshEnabled: Boolean = true
     set(value) {
       field = value
-      setPullToRefresh(value)
+      setPullToReload(value)
```
pfeiffer, line 43: Let's change that back.

packages/turbo/android/src/main/java/com/reactnativeturbowebview/RNVisitableView.kt
```diff
@@ -73,8 +74,8 @@ class RNVisitableView(context: Context) : LinearLayout(context), SessionSubscrib
     addView(visitableView)
 
     turboView.apply {
-      initializePullToRefresh(this)
-      initializeErrorPullToRefresh(this)
+      initializePullToReload(this)
+      initializeErrorPullToReload(this)
```
pfeiffer, line 78: Let's change that back.

packages/turbo/android/src/main/java/com/reactnativeturbowebview/RNVisitableView.kt
```diff
@@ -235,7 +247,7 @@ class RNVisitableView(context: Context) : LinearLayout(context), SessionSubscrib
     turboView.removeErrorView()
   }
 
-  private fun setPullToRefresh(enabled: Boolean) {
+  private fun setPullToReload(enabled: Boolean) {
```
pfeiffer, line 250: Let's change that back.

packages/turbo/android/src/main/java/com/reactnativeturbowebview/RNVisitableView.kt
```diff
@@ -294,12 +306,12 @@ class RNVisitableView(context: Context) : LinearLayout(context), SessionSubscrib
 
   override fun onZoomed(newScale: Float) {
     currentlyZoomed = true
-    setPullToRefresh(false)
+    setPullToReload(false)
```
pfeiffer, line 309: Let's change that back.

packages/turbo/android/src/main/java/com/reactnativeturbowebview/RNVisitableView.kt
```diff
@@ -294,12 +306,12 @@ class RNVisitableView(context: Context) : LinearLayout(context), SessionSubscrib
...
   }
 
   override fun onZoomReset(newScale: Float) {
     currentlyZoomed = false
-    setPullToRefresh(pullToRefreshEnabled)
+    setPullToReload(pullToRefreshEnabled)
```
pfeiffer, line 314: Let's change that back.

PR #119, opened 2024-02-14
Refactor RNVisitableViewController

pklatka:


packages/turbo/ios/RNVisitableViewController.swift
```diff
@@ -24,36 +24,82 @@ public protocol RNVisitableViewControllerDelegate {
...
+
+  open weak var visitableDelegate: VisitableDelegate?
+  open var visitableURL: URL!
+
+  private var reactViewController: UIViewController? = nil
+
+  public convenience init(url: URL) {
+    self.init()
+    self.visitableURL = url
+  }
```
pfeiffer, line 38: Is this needed?

packages/turbo/ios/RNVisitableViewController.swift
```diff
@@ -24,36 +24,82 @@ public protocol RNVisitableViewControllerDelegate {
...
+    self.init()
+    self.reactViewController = reactViewController
+    self.visitableViewControllerDelegate = visitableViewControllerDelegate
+  }
+
+  // MARK: View Lifecycle
+
+  override func viewDidLoad() {
+    super.viewDidLoad()
+    view.backgroundColor = UIColor.white
```
pfeiffer, line 50: Is this needed?

packages/turbo/ios/RNVisitableViewController.swift
pklatka: We can remove that

packages/turbo/ios/RNVisitableViewController.swift
pklatka: If we remove that we won't attach the webview :D

packages/turbo/ios/RNVisitableViewController.swift
```diff
@@ -24,36 +24,82 @@ public protocol RNVisitableViewControllerDelegate {
...
+    installVisitableView()
+  }
+
   override func viewWillAppear(_ animated: Bool) {
     super.viewWillAppear(animated)
-    delegate?.visitableWillAppear(visitable: self)
+    visitableViewControllerDelegate?.visitableWillAppear(visitable: self)
```
pfeiffer, line 51: Hm. I think we'll be missing some delegations here, right?

Previously we had _2_ delegates being called here; `delegate` and then we called `super` as well, which pointed to the `VisitableViewController`.

In `VisitableViewController` there was:

```
super.viewDidAppear(animated)
visitableDelegate?.visitableViewDidAppear(self)
```

So previously we called two delegates:
- `VisitableViewController.visitableDelegate`
- `delegate`

Now we call one:
- `visitableViewControllerDelegate`

pfeiffer: I'm kinda thinking if the code here works at all - have you tested it? As I read it, no lifecycle methods would be called on `Session` which essentially breaks everything, right?

packages/turbo/ios/RNVisitableViewController.swift
```diff
@@ -24,36 +24,77 @@ public protocol RNVisitableViewControllerDelegate {
...
+    visitableViewControllerDelegate?.visitableDidDisappear(visitable: self)
   }
-  
-  override func showVisitableActivityIndicator(){
-    delegate?.showVisitableActivityIndicator()
+
+  // MARK: Visitable
+
+  func visitableDidRender() {
+    title = visitableView.webView?.title
```
pfeiffer, line 66: Not needed?

packages/turbo/ios/RNVisitableViewController.swift
```diff
@@ -24,36 +24,77 @@ public protocol RNVisitableViewControllerDelegate {
...
+  func hideVisitableActivityIndicator() {
+    visitableViewControllerDelegate?.hideVisitableActivityIndicator()
+  }
+    
+  // MARK: Visitable View
+
+  open private(set) lazy var visitableView: VisitableView! = {
+    let view = VisitableView(frame: CGRect.zero)
+    view.translatesAutoresizingMaskIntoConstraints = false
+    return view
```
pfeiffer, line 83:
```suggestion
    view.translatesAutoresizingMaskIntoConstraints = false
    
    return view
```

packages/turbo/ios/RNVisitableViewController.swift
pklatka: It is working because of this line of code in the RNVisitableView:
```
    didSet {
      if(url != oldValue) {
        visit()
      }
    }
  }
```

It is called on every page visit, so if we call `visitableDelegate?.visitableViewWillAppear(self)` we're gonna call `visit()` twice which we want to avoid. I think we want to revert the behavior of didSet which was introduced in #107  and let's delegate that to the visitableDelegate.

pklatka: Okay, without that we won't initialize the visit at all, `visitableViewWillAppear` handles cancelling back gestures etc so actually we shouldn't call this twice.

packages/turbo/ios/RNVisitableViewController.swift
```diff
@@ -24,36 +24,79 @@ public protocol RNVisitableViewControllerDelegate {
...
 }
 
-class RNVisitableViewController: VisitableViewController {
-  
-  public var delegate: RNVisitableViewControllerDelegate?
-  
+class RNVisitableViewController: UIViewController, Visitable {
+  public var visitableViewControllerDelegate: RNVisitableViewControllerDelegate?
+
+  open weak var visitableDelegate: VisitableDelegate?
```
pfeiffer, line 30: I think we get this delegate from `Visitable` protocol automatically?

packages/turbo/ios/RNVisitableViewController.swift
```diff
@@ -24,36 +24,79 @@ public protocol RNVisitableViewControllerDelegate {
   
 }
 
-class RNVisitableViewController: VisitableViewController {
-  
-  public var delegate: RNVisitableViewControllerDelegate?
-  
+class RNVisitableViewController: UIViewController, Visitable {
+  public var visitableViewControllerDelegate: RNVisitableViewControllerDelegate?
```
pfeiffer, line 28: Any reason for renaming this from `delegate`?

packages/turbo/ios/RNVisitableViewController.swift
pfeiffer: I mean the background color

packages/turbo/ios/RNVisitableViewController.swift
pklatka: No, we don't
image

PR #121, opened 2024-02-15
Compare significant params only

pklatka:
## Summary

This PR adds a `areParamsSimilar` function to compare the params of 2 screens and see if they are to be considered “similar” enough to be removed. For this, we ignore path (which is always different, even if we should replace!) and params (which are also different).


pfeiffer:
LGTM

PR #130, opened 2024-02-26
Manually call `beginAppearanceTransition` when controller is contained within a `UIPageViewController`

pklatka:
## Summary
This PR:
1. Manually calls `beginAppearanceTransition` when controller is contained within a `UIPageViewController`. Unfortunately sometimes `viewWillAppear` is not being called when first tab in `material-top-tabs` is mounted. We call `beginAppearanceTransition` to make sure that we notify the `RNVisitableViewController`.
2.  Changes code order in `didMoveToWindow` method.


pfeiffer:
LGTM and seems to have fixed the issue with tab views.

PR #131, opened 2024-02-26
iOS: Make `controller` property optional in `RNVisitableView`

pklatka:
## Summary

This PR makes the `controller` property optional in `RNVisitableView` on iOS. The reason for this change is to handle a case where we call `refreshSession` on a session that has no mounted native component. Currently, because we set controller property to `nil`, this results in an app crash.

### DIFF
--- a/packages/turbo/ios/RNVisitableView.swift
+++ b/packages/turbo/ios/RNVisitableView.swift
@@ -48,8 +48,13 @@ class RNVisitableView: UIView, RNSessionSubscriber {
     return configuration
   }()
     
-  lazy var _controller: RNVisitableViewController? =  RNVisitableViewController(reactViewController: reactViewController(), delegate: self)
-  var controller: RNVisitableViewController { _controller! }
+  var _controller: RNVisitableViewController? = nil
+  var controller: RNVisitableViewController {
+    if (_controller == nil){
+      _controller = RNVisitableViewController(reactViewController: reactViewController(), delegate: self)
+    }
+    return _controller!
+  }
     
   private var isRefreshing: Bool {
     controller.visitableView.isRefreshing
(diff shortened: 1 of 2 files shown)


packages/turbo/ios/RNVisitableView.swift
```diff
@@ -97,6 +102,7 @@ class RNVisitableView: UIView, RNSessionSubscriber {
   }
 
   public func refresh() {
+    controller.visitableURL = URL(string: String(url))
     session.visit(controller, action: .replace)
```
pfeiffer, line 106: Hm. I don't think we want to support refreshing if the controller is `nil`; we'd be unmounted, right?

pfeiffer: Let's add a guard instead and not do anything when the controller is not there.

Also let's make the `controller` optional and remove the `_controller` prop.

packages/turbo/ios/RNVisitableView.swift
```diff
@@ -119,19 +118,20 @@ class RNVisitableView: UIView, RNSessionSubscriber {
   }
 
   public func refresh() {
-    session.visit(controller, action: .replace)
+    guard controller != nil else { return }
```
pfeiffer, line 121: Let's do flow style to keep it consistent.

packages/turbo/ios/RNVisitableView.swift
```diff
@@ -75,22 +74,22 @@ class RNVisitableView: UIView, RNSessionSubscriber {
     guard window != nil else { return }
```
pfeiffer, line 74: Let's use flow

packages/turbo/ios/RNVisitableView.swift
```diff
@@ -119,19 +118,20 @@ class RNVisitableView: UIView, RNSessionSubscriber {
...
 
   private func visit() {
-    if (controller.visitableURL?.absoluteString == url as String) {
+    if (controller?.visitableURL?.absoluteString == url as String) {
       return
     }
     performVisit()
   }
 
   private func performVisit() {
```
pfeiffer, line 132: We can move the implementation here to `visit`

PR #135, opened 2024-02-28
Use cloned `turbo-ios` repository as a library source

pklatka:
## Summary

This PR adds a script which clones `turbo-ios` library when installing a pod.  This change is necessary to easily use the latest releases of `turbo-ios`.


patches/turbo-js-resource.patch
pfeiffer, line 1: There's no way we can avoid this patch? 😭

scripts/update-turbo-ios.sh
```diff
@@ -0,0 +1,32 @@
...
+    exit 1
+fi
+
+# Patch applied to turbo-ios which fixes the resource loading issue
+PATCH_PATH=$(realpath ../patches/turbo-js-resource.patch)
+
+# Update the turbo-ios git submodule
+cd ../packages/turbo/ios/turbo-ios
+git fetch --all --tags --prune
+git checkout tags/$1 -B $1
```
pfeiffer, line 14: Wondering if we should prefix the new branch to be clear that we're using on a "patched" version of the tag?

pfeiffer: Hm. One issue I'm thinking of here is that the ´ref` that we're pointing to in the submodule, will be non-existent in the `turbo-ios` library without no way to figure which version we're pointing to.

packages/turbo/ios/vendor/scripts/install-turbo-ios.sh
```diff
@@ -0,0 +1,13 @@
...
+if [ -z "$1" ]
+  then
+    echo "No version tag supplied"
+    exit 1
+fi
+
+# Shallow clone the turbo-ios repo
+cd ./ios/vendor
+rm -rf turbo-ios
+git clone --branch $1 --depth 1 $TURBO_IOS_REPO_PATH
```
pklatka, line 13: I'm thinking that we should probably include `.framework` file and not clone the repo when building the app, just use `turbo-ios` as a "local lib".

pklatka:
Moved to #140

PR #136, opened 2024-02-29
Remove backgroundColor from default error and loading views

cspeer:
I removed the `backgroundColor` from the error and loading view wrappers. This causes the views to be transparent and therefore it won't show a white loading screen when the color scheme is dark or otherwise incompatible with the color scheme of the app.

### DIFF
--- a/packages/turbo/src/hooks/useWebViewState.tsx
+++ b/packages/turbo/src/hooks/useWebViewState.tsx
@@ -100,7 +100,6 @@ const styles = StyleSheet.create({
     alignItems: 'center',
     height: '100%',
     width: '100%',
-    backgroundColor: 'white',
     zIndex: 1,
   },
   title: {


pklatka:
Hi @cspeer,

Thanks for helping improve our library!

I’m not sure if this change makes sense - you can always use `renderError` or `renderLoading` prop to define your custom error/loading component. This approach of defining a custom error/loading component is similar to `react-native-webview` approach - they also show a custom error/loading component with a [white background](https://github.com/react-native-webview/react-native-webview/blob/master/src/WebView.styles.ts#L17-L25). In my opinion, this component doesn’t look nice with a background color different from white. Also, the text color will always remain black, and setting default app color to black would make this text unreadable.

cspeer:
Hi @pklatka,

yeah, I totally get that and you're absolutely right about the text color. I didn't think of that.
What bugs me though is that you have to write your own, or at least copy, the default Error and LoadingViews just to have them conform to your color scheme.
I think a package like this shouldn't impose any colors or at least be agnostic about it.
So, how about we keep the color property in the stylesheet but make it customizable by either allowing the `WebView` to pass the style prop and then merge it, or the `WebView` could expose 2 specific properties for text and background color. 
What do you think?

pklatka:
Hi, sorry for a late response. Adding an additional prop to `VisitableView` to change the style of the error/loading component is not a good idea. This way, we make `VisitableView` _too responsible_ for everything. Also, I'm pretty sure that not many people would use that prop. Therefore, the right approach is to define `renderError` and `renderLoading` props.

Sadly, I need to close this PR, but thank you for spending your precious time on improving our library!

PR #137, opened 2024-03-04
Use `turbo-android` as a local library

pklatka:
## Summary

This PR adds a script which generates `.aar` file for `turbo-android` library.  This change is necessary to easily use the latest releases of `turbo-android`.


pklatka:
Moved to #140

PR #144, opened 2024-03-16
Avoid calling the native method before the native `sessionHandle` prop is set up

pklatka:
## Summary

This PR 
1. Wraps `dispatchViewManagerCommand` function with `setTimeout`. This way we avoid calling the native method before the native `sessionHandle` prop is set up. 
2. Introduces `nextEventLoopTick` function whose purpose is to map the behavior of [process.nextTick](https://nodejs.org/docs/latest/api/process.html#processnexttickcallback-args) function.


pfeiffer:
LGTM.

PR #158, opened 2024-03-20
Fix issue with routes in `getStateForPath`

pfeiffer:
This PR fixes a bug where the `fullPath` param was always set on the first route in the unpacked state. The first route might not always be the active one, so this PR changes it to use the states `index` to determine the active route in the `routes`.

### DIFF
--- a/packages/navigation/src/buildWebScreen.tsx
+++ b/packages/navigation/src/buildWebScreen.tsx
@@ -1,28 +1,28 @@
 import {
-  PartialRoute,
-  Route,
   getStateFromPath,
+  PartialState,
 } from '@react-navigation/native';
+import type { NavigationState } from '@react-navigation/core';
 
 import { LinkingConfig } from './hooks/useCurrentUrl';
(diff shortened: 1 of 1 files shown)


packages/navigation/src/buildWebScreen.tsx
```diff
@@ -1,28 +1,28 @@
...
 
 function getParams(
-  routes: PartialRoutes | undefined
+  state: NavigationState | PartialState<NavigationState>
 ): LinkedParams | undefined {
-  const firstRoute = routes?.[0];
-  if (firstRoute) {
-    if (firstRoute.params) {
-      return firstRoute.params;
+  const activeRoute = state.routes?.[state.index ?? 0];
```
pfeiffer, line 16: This is the prominent change; previously we always returned the params for the _first_ route, not the active route. This PR changes so we use `state.index` to determine the active route.

pklatka (approved):
Good catch 👍

PR #162, opened 2024-03-21
Android: override default value of empty `<title>` tag

pklatka:
## Summary

When `<title>` tag is empty, native WebView on Android sets the title to `host + path`. That's a different behavior from `WKWebView` on iOS. This PR makes sure that we get the same title result.

Note: if someone sets the title to the host (e.g., swmansion.com), it will be set to an empty string. In this case, to use a `host + path` combination as a title, it should be set in the `options` parameter in `react-navigation`.

## Test plan

Set the title tag to empty string and check if value is correct.


pfeiffer:
What about just doing it in JS land?

We have `handleOnLoad` with access to both `title` and `url`

pklatka:
Closing, decided to add a comment mentioning this behavior: #165

PR #172, opened 2024-03-27
Change logic of `refresh` method

pklatka:
## Summary

This PR changes the logic of `refresh` method - we no longer dispatch `replace` visit, instead, the JS script is injected which calls `refresh()` method on `Turbo.session` object. 

## Test plan

Added a `setTimeout` in `WebView.tsx` and checked if website is refreshed properly.

### DIFF
--- a/packages/turbo/android/src/main/java/com/reactnativeturbowebview/RNVisitableView.kt
+++ b/packages/turbo/android/src/main/java/com/reactnativeturbowebview/RNVisitableView.kt
@@ -24,8 +24,8 @@ import dev.hotwire.turbo.errors.WebSslError
 import dev.hotwire.turbo.visit.TurboVisitAction
 
 const val REFRESH_SCRIPT = "typeof Turbo.session.refresh === 'function'" +
-        "? Turbo.session.refresh(document.baseURI)" +
-        ": Turbo.visit(document.baseURI, { action: 'replace', shouldCacheSnapshot: 'false' })"
+        "? Turbo.session.refresh(document.baseURI)" + // Turbo 8+
+        ": Turbo.visit(document.baseURI, { action: 'replace', shouldCacheSnapshot: 'false' })" // Older Turbo versions 
 
 class RNVisitableView(context: Context) : LinearLayout(context), SessionSubscriber {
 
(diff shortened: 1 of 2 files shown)


packages/turbo/android/src/main/java/com/reactnativeturbowebview/RNVisitableView.kt
```diff
@@ -23,6 +23,10 @@ import dev.hotwire.turbo.errors.WebError
 import dev.hotwire.turbo.errors.WebSslError
 import dev.hotwire.turbo.visit.TurboVisitAction
 
+const val REFRESH_SCRIPT = "typeof Turbo.session.refresh === 'function'" +
+        "? Turbo.session.refresh(document.baseURI)" +
+        ": Turbo.visit(document.baseURI, { action: 'replace', shouldCacheSnapshot: 'false' })"
```
pfeiffer, line 28:
```suggestion
const val REFRESH_SCRIPT = "typeof Turbo.session.refresh === 'function'" +
        "? Turbo.session.refresh(document.baseURI)" + // Turbo 8+
        ": Turbo.visit(document.baseURI, { action: 'replace', shouldCacheSnapshot: 'false' })" // Older Turbo versions 
```

packages/turbo/ios/RNVisitableView.swift
```diff
@@ -7,6 +7,10 @@
 
 import UIKit
 
+let REFRESH_SCRIPT = "typeof Turbo.session.refresh === 'function'" +
+            "? Turbo.session.refresh(document.baseURI)" +
+            ": Turbo.visit(document.baseURI, { action: 'replace', shouldCacheSnapshot: 'false' })"
```
pfeiffer, line 12:
```suggestion
let REFRESH_SCRIPT = "typeof Turbo.session.refresh === 'function'" +
            "? Turbo.session.refresh(document.baseURI)" + // Turbo 8+
            ": Turbo.visit(document.baseURI, { action: 'replace', shouldCacheSnapshot: 'false' })" // Older Turbo versions
```

pfeiffer:
LGTM!

PR #183, opened 2024-04-12
Ensure that `RNVisitableViewController` is always defined

pklatka:
## Summary

This PR refactors the way of using the `RNVisitableViewController` in `RNVisitableView` - now every time the controller variable is accessed, it is caught by a proxy which checks if the controller is `nil` and takes appropriate action as needed.

## Test plan


pfeiffer:
Under what circumstances could the `controller` be `nil` and the need for this new proxy be effectuated?

pklatka:
Closing for now, need to investigate the issue more.

PR #184, opened 2024-04-16
Android: ensure that `WebView` is detached when application context is available

pklatka:
## Summary

This PR adds a check in `onDropViewInstance` which ensures that `detachWebView` is called when the app context is available. A similar check is done in `ViewManager.onDropViewInstance` when recycling the view.


pfeiffer:
LGTM.

PR #190, opened 2024-04-19
Introduce `scrollEnabled` property, refactor a way of setting the `WebView` settings, and apply minor fixes to the native codebase

pklatka:
## Summary

This PR:
1. Implements `scrollEnabled` property
2. Modifies the approach of applying the settings in the WebView
3. Applies minor fixes to the native codebase

## Test plan

Tested if dynamically changing the `scrollEnabled` works properly in the demo app:
1. Tested if the prop is updated on the native side when on the same webpage (e.g. setTimeout which changes the value)
2. Tested if the changes are applied when going back to the previous screen


packages/turbo/ios/RNSessionConfiguration.swift
```diff
@@ -0,0 +1,27 @@
+//
+//  RNSessionConfiguration.swift
+//  RNTurbo
+//
+//  Created by Patryk Klatka on 22/04/2024.
+//
+
+import Foundation
+
+class RNSessionConfiguration {
```
pfeiffer, line 10: Something bugs me with the naming. Are we configuring a Session? I don't think so.

packages/turbo/ios/RNSessionConfiguration.swift
```diff
@@ -0,0 +1,27 @@
...
+  var sessionDelegate: RNSession? = nil {
```
pfeiffer, line 11: `sessionDelegate`? It's not really a session _delegate_ here, right? It's the actual `session`

packages/turbo/ios/RNVisitableView.swift
```diff
@@ -14,14 +14,23 @@ let REFRESH_SCRIPT = "typeof Turbo.session.refresh === 'function'" +
...
   @objc var url: NSString = "" {
     didSet {
       if(url != oldValue) {
         visit()
       }
     }
   }
+  @objc var applicationNameForUserAgent: NSString? = nil {
+    didSet {
+      sessionConfiguration.webViewConfiguration.applicationNameForUserAgent = applicationNameForUserAgent as? String
```
pfeiffer, line 26: If the pattern with the configurator should be there, I think _it_ should have a `applicationNameForUserAgent` property that we set directly here. It feels a bit off to access (and manipulate) the deeply nested `applicationNameForUserAgent` and it breaks the pattern of having the configurator object.

Eg. something like:

```swift
didSet {
  sessionConfiguration.applicationNameForUserAgent =  ..
}
```

and then in the configurator:

```swift
var applicationNameForUserAgent {
  didSet {
    webViewConfiguration.applicationNameForUserAgent = applicationNameForUserAgent
  }
}
```

packages/turbo/ios/RNVisitableView.swift
```diff
@@ -14,14 +14,23 @@ let REFRESH_SCRIPT = "typeof Turbo.session.refresh === 'function'" +
...
+    }
+  }
+  @objc var scrollEnabled: Bool = true {
+    didSet {
+      sessionConfiguration.isScrollEnabled = scrollEnabled
+    }
+  }
   @objc var pullToRefreshEnabled: Bool = true {
     didSet {
       controller!.visitableView.allowsPullToRefresh = pullToRefreshEnabled
```
pfeiffer, line 36: I feel like this could also be moved into the configurator provided we initialize the configurator with the `RNVisitableView` (which would allow us to access both session and controller).

packages/turbo/ios/RNSessionConfiguration.swift
```diff
@@ -0,0 +1,27 @@
...
+
+  var webViewConfiguration: WKWebViewConfiguration = WKWebViewConfiguration()
+  var isScrollEnabled: Bool = true {
+    didSet {
+      updateConfiguration()
+    }
+  }
+    
+  func updateConfiguration() {
+    sessionDelegate?.webView.scrollView.isScrollEnabled = isScrollEnabled
```
pfeiffer, line 25: I'm wondering if the access to the scroll view via session is needed at all?

Couldn't we just do:

```
controller!.visitableView.webView?.isScrollEnabled = ...
```

pfeiffer:
Had a look at at least the iOS code and I'm wondering if a slightly different approach could be taken:

```swift
@objc var scrollEnabled {
  didSet {
    webView?.isScrollEnabled = scrollEnabled
  }
}

...

func visitableDidAppear(..) // or visitableDidRender? {
  webView?.isScrollEnabled = scrollEnabled
}
```

Could even be refactored to a method, eg `configureWebView` if we have a lot of options that need to be set on the WebView, eg.:

```swift
@objc var scrollEnabled {
  didSet {
     configureWebView()
  }
}

@objc var contentInset {
  didSet {
     configureWebView()
  }
}
...

func visitableDidAppear(..) // or visitableDidRender? {
  configureWebView()
}

func configureWebView() {
   guard webView != nil else { return }

   webView.isScrollEnabled = scrollEnabled
   webView.contentInset = contentInset
   ...
}

File: packages/turbo/ios/RNVisitableView.swift [hunk 6]
```diff
@@ -52,13 +52,19 @@ class RNVisitableView: UIView, RNSessionSubscriber {
...
-    return session
-  }()
-  private lazy var webView: WKWebView = session.webView
+  private var _session: RNSession? = nil
+  private var session: RNSession? {
+    if (_session != nil) {
+      return _session
+    }
+      
+    if (sessionHandle == nil){
```
pfeiffer, line 61: Missing space

packages/turbo/android/src/main/java/com/reactnativeturbowebview/RNVisitableView.kt
```diff
@@ -92,8 +106,30 @@ class RNVisitableView(context: Context) : LinearLayout(context), SessionSubscrib
     }
   }
 
+  private fun updateWebViewConfiguration() {
+    if (webView == null) return
+    setUserAgentString(webView!!, applicationNameForUserAgent)
+    setOnTouchListener(webView!!, scrollEnabled)
```
pfeiffer, line 112:
```suggestion
  private fun updateWebViewConfiguration() {
    if (webView == null) return
    
    setUserAgentString(webView!!, applicationNameForUserAgent)
    setOnTouchListener(webView!!, scrollEnabled)
```

packages/turbo/android/src/main/java/com/reactnativeturbowebview/RNSession.kt
```diff
@@ -38,7 +36,6 @@ class RNSession(
     WebView.setWebContentsDebuggingEnabled(BuildConfig.DEBUG)
     webView.settings.setJavaScriptEnabled(true)
     webView.addJavascriptInterface(JavaScriptInterface(), "AndroidInterface")
-    setUserAgentString(webView, applicationNameForUserAgent)
```
pfeiffer, line 41: What's the reason for moving this into the configure WebView flow?

I'm primarily thinking that iOS doesn't support updating the user agent after the initialization of a session and wondering if it's better to keep the behavior consistent between platforms?

PR #192, opened 2024-04-22
turbo-ios: remove `UIRefreshControl` layout constraints

pklatka:
## Summary

This PR adds a patch for `turbo-ios` library which removes the layout constraints which are set in `VisitableView.installRefreshControl` method. This unfortunately is necessary to make the `UIRefreshControl` work properly with `contentInset` property.

## Test plan


pfeiffer:
LGTM

PR #194, opened 2024-04-22
iOS: ensure that all completion handlers have been called

pklatka:
## Summary

This PR makes sure that all completion handlers have been called before webView navigates to other url. There is a possibility that `sendAlertResult` or `sendConfirmResult` are going to be called too late which can result in `NSInternalInconsistencyException`.


pfeiffer:
LGTM

PR #203, opened 2024-04-29
iOS: introduce `refreshControlTopAnchor` property

pklatka:
## Summary

This PR introduces a `refreshControlTopAnchor` property which allow to set the topAnchor of the native refresh control on iOS. Removing the constraint completely can introduce some unexpected behaviors, especially when using transparent header, so with this prop we can manually set the desired top offset of the native refresh control when needed.

## Test plan


pklatka:
Closing, the issue with wrongly positioned `UIRefreshControl` can be solved by setting the `contentInset` property

PR #217, opened 2025-03-28
chore: Update expo-sdk to 52

sztobar:
## Summary

* update expo-sdk to 52 (verified backwards compatibility with expo 50)
* update react-native to 0.76.7 (keep the default rn version of expo and leave react on v18). For more info check expo [changelog](https://expo.dev/changelog/react-native-78).
* updated react-navigation to v7 (verified backwards compatibility with react-navigation v6)
* update all turbo-demo-expo-example dependencies to latest
* update all dependencies in react-native-web-screen and react-native-turbo to latest
* apply changes requires after bumping dependencies:
  * fixed swift missing imports in react-native-turbo
  * fixes kotlin incorrect type in react-native-turbo
  * bumped ios deploymentTarget to 15.6 in turbo-demo-expo-example
  * added compilerOptions.paths for react-navigation/native in react-native-web-screen

### Regressions

After update to react-navigation v7 I noticed that `back` button on ios is misaligned. I'm looking for the cause and fix for that.
Simulator Screenshot - iPhone SE (3rd generation) - 2025-03-31 at 16 17 58

## Test plan

* no new features are added, just running the example on ios and android to verify that app is working.


Fryzu (approved):
Hey @sztobar LGTM thanks for the update 🚀 
can we please make sure that this doesn't introduce any breaking changes for previous supported version of RN (from the code it seems like it doesn't)?

packages/navigation/src/hooks/useWebviewNavigate.ts
```diff
@@ -11,9 +11,9 @@ import {
   useNavigation,
   useRoute,
   CommonActions,
+  LinkingContext,
 } from '@react-navigation/native';
-import LinkingContext from '@react-navigation/native/src/LinkingContext';
-import extractPathFromURL from '@react-navigation/native/src/extractPathFromURL';
+import { extractPathFromURL } from '@react-navigation/native/src/extractPathFromURL';
```
pklatka, line 16: I’m not a fan of having to change the peer dependencies just because the export type of `extractPathFromURL` is different in v7. I think most of the apps which are built around this library are using v6 and upgrading to v7 in some extreme cases might be _unpleasant_ (at least that's the impression I get from reading the upgrade guide from v6). By continuing to support v6 in the peer dependencies, we make the library accessible to a wider range of applications that can benefit from it.

In this case, I would do something like that:

```js
import * as extractPathFromURLLib from '@react-navigation/native/src/extractPathFromURL';

// ...

// explanation comment
const extractPathFromURL =
  extractPathFromURLLib.default || extractPathFromURLLib;
```

pklatka (approved):

packages/navigation/src/hooks/useWebviewNavigate.ts
sztobar: good idea! Implemented in [39d5b20](https://github.com/software-mansion-labs/react-native-turbo-demo/pull/217/commits/39d5b20640d990dbb2b82b61713602a9dc7114cc)

packages/navigation/package.json
```diff
@@ -27,12 +27,12 @@
...
   "devDependencies": {
-    "@react-navigation/core": "^6.0.0",
+    "@react-navigation/core": "^7.7.0",
     "react-native-turbo": "*"
   },
   "peerDependencies": {
-    "@react-navigation/core": ">=6.0.0",
-    "@react-navigation/native": ">=6.0.0",
+    "@react-navigation/core": ">=7.7.0",
+    "@react-navigation/native": ">=7.0.19",
```
pfeiffer, line 35: If `react-navigation` v6 is still supported via the custom export, should the dependency requirement here be loosened again?

packages/navigation/src/hooks/useWebviewNavigate.ts
```diff
@@ -11,12 +11,12 @@ import {
   useNavigation,
   useRoute,
   CommonActions,
+  LinkingContext,
 } from '@react-navigation/native';
-import LinkingContext from '@react-navigation/native/src/LinkingContext';
```
pfeiffer, line 15: If we still want to support v6 (which I think we should), do we need to do a custom export here as well similar to `extractPathFromURL`?

sztobar: I checked that before and `LinkingContext` export changed in v7, but in both v6 and v7 it can be simply imported from `'@react-navigation/native'`, so there's no need for a custom export

packages/navigation/package.json
sztobar: ahh hood point, restored it in [7cfeedf](https://github.com/software-mansion-labs/react-native-turbo-demo/pull/217/commits/7cfeedfc06aa1f5104d6c89eb0d955967b2ec09c)

PR #220, opened 2025-04-11
feat: Migrate from turbo-(ios|android) to hotwire-native-(ios|android)

sztobar:
## Summary

* renamed occurrences of turbo to hotwire in native packages, scripts
* added new API method support:
  * [visitProposedToCrossOriginRedirect](https://github.com/hotwired/hotwire-native-android/pull/82) for android
  * [didProposeVisitToCrossOriginRedirect](https://github.com/hotwired/hotwire-native-ios/pull/70) for ios
* updated screenshot handling in android implementation
* updated react-native navigation devDependency in rn package react-native-turbo to avoid bundled version mismatch
* bumped android minSdkVersion to required `28`and configured kotlinVersion to required `1.9.25` in app.json
* hotwire native is set to version `1.1.1` while the latest is `1.1.3`. This is because later version requires kotlin `2.1.0` which is not available in this react-native version. 
* update devDependecies versions in `react-native-turbo` package for `@react-navigation/core` and `@react-navigation/native` from `^6.0.0` to `>=6.0.0` to avoid incorrect version bundled in example app

## Test plan

* checked example-demo app for both android and ios and noticed no regressions


pklatka:
I can't run example app on a freshly-cloned repo, could you please clone the repository and check if the error is present on your side when running the example app on Android (see the screenshot)?
image

Small note: you can consider upgrading this, not necessarily in this PR
image

sztobar:
@pklatka are you certain that you've run the demo app using this branch? I double-checked and I noticed that this issue happens on `main` branch, this branch fixes the issue by updating react-navigation devDependencies in packages from `^6` to `>=6`. Because this branch is the next invevitable upgrade I haven't created separate PR to fix that and decide to include this here.

Regarding turbo - thanks for the heads up, I'll create separate PR to fix that. I wanted to do some cleaning with peerDependencies, there are a lot of warning when installing dependencies.

pklatka (approved):
Double checked and indeed the app works on Android, my bad. Code looks ok!

pfeiffer:
Completed QA of this in our app and found no issues. Great!

PR #222, opened 2025-04-24
Update project dependencies

sztobar:
## Summary

* update turbo from `1.13.3` to `2.5.0`
* rename example demo app scripts:
  * `lint:check` into `lint`
  * `tsc:check` into `typescript`
  to match `packages` scripts. Now running `yarn lint` or `yarn typescript` will run it for both packages and example app as well.
* added `strict: true` to example tsconfig.json for better type-checking and fixed issues that were caused by introducing strict mode.
* fix turbo postinstall script that run from `examples/turbo-demo-expo-example` on consecutive `yarn install` when turbo cache misses.

## Test plan

* built both apps and verified that there's no regressions


packages/turbo/scripts/postinstall.sh
```diff
@@ -5,15 +5,23 @@ if [ ! -d "./android/vendor" ]; then
...
 fi
 
-VENDOR_DIR=$(realpath ./android/vendor)
+CURRENT_DIR=$(realpath .)
+MONOREPO_SUFFIX="examples/turbo-demo-expo-example/node_modules/react-native-turbo"
+SKIP_RSYNC=0
 PARENT_FOLDER=$(realpath .. | awk -F/ '{print $NF}')
 
-if [ "$PARENT_FOLDER" = "node_modules" ]; then
+# skip postinstall for `yarn install` called from inside `turbo-demo-expo-example`
```
pklatka, line 13: NIT - other comments start with uppercase letter 
```suggestion
# Skip postinstall for `yarn install` called from inside `turbo-demo-expo-example`
```

pklatka (approved):

PR #224, opened 2025-04-29
feat: Update hotwire-native-ios to 1.2

sztobar:
## Summary

* update API changes that includes updating `visitableURL` to `initialVisitableURL` and `currentVisitableURL`. Related hotwire-native-ios PR: https://github.com/hotwired/hotwire-native-ios/pull/108
* introduce new method `.session(decidePolicyFor)` that opens external URL. We might consider making it more configurable with props but it's out of the scope of this PR. Related hotwire-native-ios PR: https://github.com/hotwired/hotwire-native-ios/pull/105 

## Test plan

* checked current demo web page and found no regressions. It doesn't provider bug with url change inside webview that this PR fixes, so it has to be tested on a different app

### DIFF
--- a/packages/turbo/ios/RNSession.swift
+++ b/packages/turbo/ios/RNSession.swift
@@ -116,6 +116,15 @@ extension RNSession: SessionDelegate {
   func sessionDidFinishFormSubmission(_ session: Session) {
     visitableView?.didFinishFormSubmission()
   }
+
+  func session(_ session: Session, decidePolicyFor navigationAction: WKNavigationAction) -> WebViewPolicyManager.Decision {
+    guard let url = navigationAction.request.url else {
+      return .allow
+    }
+    // regardless of the return value here nothing happens, so we have to manually open external URL
+    visitableView?.didOpenExternalUrl(url: url)
+    return .allow
+  }
 }
 
 extension RNSession: WKScriptMessageHandler {
--- a/packages/turbo/ios/RNVisitableView.swift
+++ b/packages/turbo/ios/RNVisitableView.swift
@@ -175,10 +175,10 @@ class RNVisitableView: UIView, RNSessionSubscriber {
   }
 
   private func visit() {
-    if (controller?.visitableURL?.absoluteString == url as String) {
+    if (controller?.currentVisitableURL.absoluteString == url as String) {
       return
     }
-    controller!.visitableURL = URL(string: String(url))
+    controller!.initializeVisit(url: URL(string: String(url))!)
     session?.visit(controller!)
   }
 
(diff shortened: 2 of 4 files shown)


packages/turbo/ios/RNSession.swift
```diff
@@ -116,6 +116,15 @@ extension RNSession: SessionDelegate {
...
   }
+
+  func session(_ session: Session, decidePolicyFor navigationAction: WKNavigationAction) -> WebViewPolicyManager.Decision {
+    guard let url = navigationAction.request.url else {
+      return .allow
+    }
+    // regardless of the return value here nothing happens, so we have to manually open external URL
+    visitableView?.didOpenExternalUrl(url: url)
+    return .allow
+  }
```
pfeiffer, line 127: This has introduced a new behavior in our app. This seems to be called for on-page `<iframe src="..">` elements.

As an example, in our app we use Google Tag Manager. They have an initialization script that includes an `<iframe src="https://www.googletagmanager.com/ns.html?...">`. On every page view in the app, we see that `didOpenExternalUrl` is called with the Google Tag Manager URL with no way to differentiate if it was a user-decided navigation action or an on-page iframe.

pfeiffer: Hotwire Native has an extension that provides some additional on if the navigationAction happened in the main frame etc: https://github.com/hotwired/hotwire-native-ios/blob/f18b87938f55cbe8976e96702b5512fa9afcd674/Source/Turbo/Navigator/Extensions/WKNavigationAction%2BUtils.swift#L1C1-L12C1

I think there's a few things going on here:

1. The previous `Session#openExternalURL` that we relied on to trigger the `didOpenExternalUrl` on the RN-side was changed in Hotwire Native iOS to support the cross-origin redirects.
2. We have not yet fully supported the change
3. The current implementation in this PR is to naive and triggers even for iframes etc.

pfeiffer: Opened a PR #225 that addresses this.

pfeiffer:
After extensive testing, I think this isn't ready to merge. I'll see if I can wrap up a PR that addresses the issues separately - there are a few regressions here besides the one in #225.

piaccho:
Hello @pfeiffer 👋 I was planning to merge [#225](https://github.com/software-mansion-labs/react-native-turbo/pull/225) and wanted to check if this PR fully addresses the three issues you mentioned [there](https://github.com/software-mansion-labs/react-native-turbo/pull/224#discussion_r2070069424). Regarding the unresolved regressions – could you clarify what they are exactly? If possible, a minimal, reproducible example would be super helpful. I've just started working with this repository and am trying to understand the context.

PR #227, opened 2025-05-14
Observe URL changes due to redirects and/or history changes 

pfeiffer:
## Summary

For Strada Bridge Components, we ignore messages being sent unless the `url` match where the Strada Component was originally mounted. `url` is a controlled prop on the React Native side.

This causes issues when a Turbo Frame has changed the `document.location` of the WebView to something else (ie. with `data-turbo-action="replace"`), as the `document.location` is not out-of-sync with the `url` prop on the RN side, which is used to detect if messages sent via Strada bridge should be ignored or taken action on.

In order to make sure that we know the current URL of the WebView in React Native land, we can observe the `webView.url` value on the native side, and let the RN side know that it has changed due to eg. history pushes or redirects.

## Test plan

In order to test it, we need to have a Turbo Frame with `data-turbo-action="replace"` as well as a Strada component on the replaced content page.

I've created a branch on the demo web app that contains this: https://github.com/pfeiffer/turbo-native-demo/tree/test-turbo-frame-redirects

### Before

https://github.com/user-attachments/assets/588eb3e1-3fa1-4e7c-9b04-52c2eb52f8dc

### After

https://github.com/user-attachments/assets/597f7e7e-20d5-4e44-bb14-e93b8bdab608

### DIFF
--- a/packages/turbo/ios/RNVisitableView.swift
+++ b/packages/turbo/ios/RNVisitableView.swift
@@ -58,6 +58,7 @@ class RNVisitableView: UIView, RNSessionSubscriber {
   @objc var onShowLoading: RCTDirectEventBlock?
   @objc var onHideLoading: RCTDirectEventBlock?
   @objc var onContentProcessDidTerminate: RCTDirectEventBlock?
+  @objc var onUrlChange: RCTDirectEventBlock?
 
   private var onConfirmHandler: ((Bool) -> Void)?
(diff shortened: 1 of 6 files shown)


pfeiffer:
It was not easy to do something similar for Android, so I've opened #228 instead which has a slightly different approach but works on both Android and iOS.

PR #228, opened 2025-05-14
Fix BridgeComponents after Turbo Frame navigations or redirects

pfeiffer:
## Summary

This PR is an alternative to #227 that fixes issues with Strada Bridge Components as URLs change. It also fixes a bug in the existing code, where message listeners were not removed as bridge components were unmounted. The PR here is universal and works for both Android and iOS.

Strada Bridge Components compares the `url` it was initialized with the `url` of the metadata from the WebView. This also means that if any of these become out-of-sync, Strada components will stop working. This out-of-sync can happen when eg. the WebView's history changes with a Turbo Frame navigation.

In #227, a solution to this was to observe the native WebView's `url` and sync changes back to RN land. This works fine in iOS, but requires a lot of hacks to work on Android.

This PR fixes it in a different way by removing the `url` matching in Strada Components.

We originally introduced the `url` match assertion based on the Turbo Native iOS code. However, the implementation in React Native vs. Turbo Native iOS differ significantly; in React Native, we mount Bridge Components _with a key based on the url_, while in Turbo Native iOS, the bridge components are not removed when changing URL.

This means that in React Native, we can actually rely on the components being unmounted as navigation happens and the url changes. A new component will be mounted instead on the new url.

Based on this pattern, we can remove the `url` matching and assume that any messages received by the component with a matching name is related to _that_ url.

During my testing I saw that Strada Components were not being properly removed from the message listener queue when unmounted. This has also been addressed here.

## Test plan

As with #227, we can test that the Strada components work as expected by using the specific branch of the demo server.

I've created a branch on the demo web app that contains this: https://github.com/pfeiffer/turbo-native-demo/tree/test-turbo-frame-redirects

### DIFF
--- a/packages/turbo/src/BridgeComponent.ts
+++ b/packages/turbo/src/BridgeComponent.ts
@@ -5,14 +5,12 @@ import type {
   StradaMessage,
   StradaMessages,
   StradaComponentProps,
+  SessionMessageCallback,
 } from './types';
 
 const stradaMessageListener = (component: BridgeComponent) => (e: object) => {
   const message = e as StradaMessage;
-  if (
-    message?.component !== component.name ||
-    message?.data?.metadata?.url !== component.url
-  ) {
+  if (message?.component !== component.name) {
     return;
   }
   component.previousMessages[message.event] = message;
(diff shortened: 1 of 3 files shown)


piaccho:
Closing this PR as the issue was resolved in #230.

PR #230, opened 2025-06-17
Fix: BridgeComponents and sync iOS URL after frame navigations

piaccho:
## Summary

This PR provides a comprehensive fix for issues with Strada Bridge Components that occur after Turbo Frame navigations or redirects. It ensures that components remain active and that native iOS state remains synchronized, correcting functionality like pull-to-refresh.

This PR addresses two problems - #228 and iOS Pull-to-Refresh regression. To fix the iOS regression, this PR introduces an approach similar to what was mentioned in #227 - it has been resolved by adding a Key-Value Observer to the WKWebView's url property in RNVisitableView.swift. This restores pull-to-refresh and ensures native state consistency. 

## Test plan

We can test both Bridge components and pull-to-refresh functionalities work as expected by using the specific branch of the demo server.

This [branch](https://github.com/pfeiffer/turbo-native-demo/tree/test-turbo-frame-redirects?rgh-link-date=2025-05-14T10%3A44%3A15Z) of the demo web app includes Strada Components flow that can be used to test them.

## Demo

Functionalities tested on [this version](https://github.com/pfeiffer/turbo-native-demo/tree/test-turbo-frame-redirects?rgh-link-date=2025-05-14T10%3A44%3A15Z) of demo web app.

https://github.com/user-attachments/assets/0e45133e-bc4e-4e90-b9af-9fc2696118c6

### DIFF
--- a/packages/turbo/ios/RNVisitableView.swift
+++ b/packages/turbo/ios/RNVisitableView.swift
@@ -99,7 +99,27 @@ class RNVisitableView: UIView, RNSessionSubscriber {
                                                     bottom: contentInset["bottom"] ?? 0,
                                                     right: contentInset["right"] ?? 0)
   }
+
+  private var webViewUrlObservation: NSKeyValueObservation?
+
+  private func startObservingWebViewUrl() {
+    stopObservingWebViewUrl()
+    webViewUrlObservation = webView?.observe(\.url, options: [.new]) { [weak self] webView, change in
+            guard let self = self,
+                  let newUrl = change.newValue ?? nil,
+                  let controller = self.controller else { return }
+            
+            if controller.visitableURL != newUrl {
+                controller.visitableURL = newUrl
+            }
+        }   
+    }
     
+    private func stopObservingWebViewUrl() {
+        webViewUrlObservation?.invalidate()
+        webViewUrlObservation = nil
+    }
+
   override func willMove(toWindow newWindow: UIWindow?) {
     super.willMove(toWindow: newWindow)
     
@@ -115,6 +135,7 @@ class RNVisitableView: UIView, RNSessionSubscriber {
     super.didMoveToWindow()
     
     if (window == nil) {
+      stopObservingWebViewUrl()
       return
     }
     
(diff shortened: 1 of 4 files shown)


packages/turbo/ios/RNVisitableView.swift
```diff
@@ -99,7 +99,27 @@ class RNVisitableView: UIView, RNSessionSubscriber {
...
+  private func startObservingWebViewUrl() {
+    stopObservingWebViewUrl()
+    webViewUrlObservation = webView?.observe(\.url, options: [.new]) { [weak self] webView, change in
+            guard let self = self,
+                  let newUrl = change.newValue ?? nil,
+                  let controller = self.controller else { return }
+            
+            if controller.visitableURL != newUrl {
+                controller.visitableURL = newUrl
+            }
```
pklatka, line 114: 1. Shouldn't we notify the JS side that the url has been changed by Turbo? I guess this is fine when you do the `replace` action, but what about the `advance`? Don't know if this is even possible with Turbo. 
2. This feels hakish -> in this case there should be a small comment explaining this or a link to a GH issue.

pfeiffer: > Shouldn't we notify the JS side that the url has been changed by Turbo? I guess this is fine when you do the `replace` action, but what about the `advance`? Don't know if this is even possible with Turbo.

We could investigate if it'd make sense to trigger the `onLoad` prop on the JS-side, which could make sense.
The change here with KVO ensures `visitableURL` on the iOS is kept in-sync, so that pull-to-refresh will refresh the URL correctly after `history.push(..)` (eg. via Turbo Frame JS navigations on-site that do not go through the normal Turbo `advance` or `replace` flow)

> This` feels hakish -> in this case there should be a small comment explaining this or a link to a GH issue.

I think the solution is actually quite elegant, but agree that we probably need to add some comments as to what this does and why.

pklatka: Can be tracked as a follow-up.

PR #237, opened 2025-11-14
Support for New Architecture

pfeiffer:
This PR updates the library to require React Native 0.76+ and configures it for New Architecture compatibility via the interop layer.

## Breaking Changes
- **Minimum React Native version**: `>=0.76.0` (Expo SDK 52+)
- New Architecture must be enabled (RCT_NEW_ARCH_ENABLED=1)

### Build Configuration
- **iOS**: Removed conditional New Architecture checks, always enabled
- **Android**: Removed conditional checks, always apply React plugin with Codegen config

### JavaScript Layer
- Use `TurboModuleRegistry` instead of `NativeModules` for session manager
- Use Codegen-generated Commands for cleaner command dispatch
- Removed platform-specific command handling logic

### Message passing
- Changed message passing from JavaScript (WebView) -> native to always serialize to JSON for consistency between the platforms and to be able to generate codegen properly (ie. message is now always a string). We parse the JSON on the React Native side.

### Android Native Code
- **Fixed**: Changed events from bubbling to direct (consistency with iOS)
- String-based command routing instead of ordinal-based
- Minor cleanup: removed unnecessary empty array check in `getSessionHandles`
- Removed unused import in `RNWebChromeClient`

## Architecture Note
Native implementations remain on Paper architecture. The library runs via React Native's interop layer when New Architecture is enabled.

### DIFF
--- a/packages/turbo/.codegenconfig
+++ b/packages/turbo/.codegenconfig
@@ -1,0 +1,8 @@
+{
+  "name": "RNHotwireNative",
+  "type": "modules",
+  "jsSrcsDir": "src",
+  "android": {
+    "javaPackageName": "com.reactnativehotwirewebview"
+  }
+}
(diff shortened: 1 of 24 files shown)


sethhorsley:
This is awesome!!!

